PR #7, opened 2019-03-12
Basic page navigation

danebell:
Buttons to go forward and backward. "Last" button is not yet functional, because I'm not sure how (or whether it's possible) to start at the `n`th document without having to run queries (in the background) for every page up until the last.

### DIFF
--- a/ui/config.js
+++ b/ui/config.js
@@ -7,6 +7,8 @@ config.queryParams.odinsonQuery = "odinsonQuery";
 config.queryParams.parentQuery  = "parentQuery";
 config.queryParams.label        = "label";
 config.queryParams.commit       = "commit";
+config.queryParams.prevDoc      = "prevDoc";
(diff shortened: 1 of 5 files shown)


myedibleenso:
Awesome!  Forward and backward sound sufficient to me.

>I'm not sure how (or whether it's possible) to start at the nth document without having to run queries (in the background) for every page up until the last.

Hmm...to do that I think we'd need to reverse the ordering provided by the scorer (ex. `score * -1`), and then reverse the page's results once more before returning the values.

I don't think something like `.drop(totalHits - pageSize)` would scale well to millions of sentences...

danebell:
For now, I just hid the "Last" button. I put in an empty disabled button in its place for symmetry. If we really need it, we can worry about it later. I don't see a strong use case and was just hoping to add it for completeness.

PR #9, opened 2019-03-19
Delay TAG representation for faster rendering

danebell:
Provides functionality for presenting just plain text results (with elementary highlighting for matches) by default, greatly speeding initial rendering for the search results. Each sentence has its own `expand/collapse` button, and the page has an `expand/collapse all` button as well.

Limitations:
* Once collapsed, the TAG render is lost, so it takes just as long to render a second time.
* The parent object containing the `expand/collapse all` button does not track whether the individual sentences are expanded or not, so it is possible to have all sentences collapsed with the overall button still reading `Collapse all`.

### DIFF
--- a/backend/app/controllers/OdinsonController.scala
+++ b/backend/app/controllers/OdinsonController.scala
@@ -43,7 +43,7 @@ class OdinsonController @Inject() (system: ActorSystem, cc: ControllerComponents
 
   val vocabFile          = config[File]("odinson.compiler.dependenciesVocabulary")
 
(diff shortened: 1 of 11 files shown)


marcovzla:
is this ready to be merged?

myedibleenso:
not quite.  We were talking about switching the API calls first

danebell:
The commit above introduces changes such that `/api/search` will be used by default. When the expand-all button is pressed, the query is re-run with `/api/rich-search`. When a particular sentence's expand button is pressed, `/api/sentence` is called to retrieve the additional information for that sentence.

PR #24, opened 2019-04-06
collector

### LINKED ISSUE #20: Sorting results by document id
A minor issue we've been dealing with is that it's useful for us to get results sorted by doc id (e.g. it allows us to more quickly display "missing" extractions when working against a labeled dataset; it allows users to get a better feeling for the frequency of matches per documents; it provides for a stable order of results when new documents are added to the index.). 

We currently fetch all odinson's results and re-sort them but it would be nice if Odinson supported an option to return results sorted by doc-id (and potentially support paging by doc-id, etc.).

marcovzla:
Collector now returns results sorted by doc-id.
This collector is also more efficient when retrieving all matches of a query.
Closes #20 
Thanks for the help @robertvacareanu!


danebell:
Works for me. I see no problems.

PR #33, opened 2019-06-21
Add ability to match multiple expected results

BeckySharp:
In the patterns tsv they should be separated with a semi-colon

currently, there are no multiple expected, add as needed


marcovzla:
thanks!

PR #34, opened 2019-09-08
Added computeTotalHits configuration

mcshlain:
The current default behavior is to compute `totalHits` on each query. This PR adds a configuration option `computeTotalHits = true` (the default is the existing behavior). 

The reason for this PR is because the engine becomes unusable on very large data sets when many matches exist (wikipedia for example, with a rather general query). With this fix, by setting `computeTotalHits = false` you can get fast results without having to wait for the computation of `totalHits`.


marcovzla:
Thanks!

PR #37, opened 2019-09-09
Added flag to index in order of the external document id

mcshlain:
This PR adds a new configuration flag that influences the order in which documents are loaded during indexing. This flag enables the ability to load the documents based on the document id, which will cause it to match (in order) to the odinson id, consequentially making search queries results being ordered by the external document id.


marcovzla:
Thanks!

Issue #39, opened 2019-09-15
dependency vocabulary first entry is corrupted and not searchable  


mcshlain:
**Describe the bug**
The `dependencies.txt` file written to the index contains some additional bytes at the beginning of the file, causing the first dependency label to not be searchable. This is hard to detect because it only affects a single dependency label, and the specific label will change between runs of indexing. 

**Reproducing this behavior**
1. Go to the index directory and open the `dependencies.txt` file, and look at the dependency on the first line (it will be prefixed with some junk char/s). For example lets say it was `compound`, then run the query: `(?<p> []) <compound []`, you will get 0 results. 

2. Manually delete the junk prefix char/s on the first line of `dependencies.txt` and run the query again (after restarting the backend). This time you should see results. 

**Investigating the reason for the bug**
I believe this bug was introduced by: #30
From what I understand this happens because the file is written with the `FSDirectory` API but read with the LumAI File Utils (that uses regular streams from `java.io`)

Here is a code snippet to demonstrate this behavior:

```scala
  val config = ConfigFactory.load()
  // writing the data through FSDirectory API
  val directory  = FSDirectory.open(Paths.get(config[String]("odinson.indexDir")))
  val streamOut = directory.createOutput("test.txt", new IOContext())
  streamOut.writeString("abcd")
  streamOut.close()

  // reading the data through regular java.io API
  val stream1In = new BufferedReader(
                                new InputStreamReader(
                                     new FileInputStream(config[String]("odinson.indexDir") + "/test.txt")
                                )
                             )
  println("|" + stream1In.readLine() + "|")

  // reading the data through FSDirectory API
  val stream2In = directory.openInput("test.txt", new IOContext())
  println("|" + stream2In.readString() + "|")

```

The output is: 
```
|abcd|
|abcd|
```
On the first output line (reading using `java.io`), there is an extra character that causes the corruption. 

The solution should be to just use the same API for reading and writing, either the `FSDirectory` one or the `java.io` one. Just not mix them up.

myedibleenso:
Thanks for catching this, @mcshlain.

myedibleenso:
Closed by #40.

PR #40, opened 2019-09-18
use lucene api to read/write vocab

marcovzla:
addresses #39 
The file still has those weird bytes, but by consistently using lucene api (thanks @mcshlain) the issue is solved.


mcshlain:
awesome, thanks for the quick fix.

PR #56, opened 2020-04-05
unicode normalization

marcovzla:
Added `Normalizer` to normalize unicode strings for comparison.
Supports two levels of normalization:
- regular: NFC
- aggressive: NFKC, casefold, diacritic removal, some extra character replacements


marcovzla:
I think that the `Normalizer` may be useful beyond `Odinson`, so we may want to move it to `common` in the future, but I think it should live here for a while first.

PR #57, opened 2020-04-07
add FastNLPProcessor, and make default in conf

BeckySharp:
Added FastNLPProcessor as the default one in AnnotateText


BeckySharp:
decided not to update proc version for now

Issue #65, opened 2020-05-06
How to use the pattern matcher only?


felicitywang:
Hi,
Thanks for the great work(saw this from the AI's SPIKE-CORD demo). I'm working on information extraction(without the retrieval part) and this is very useful(your pattern matching is a lot more powerful than Spacy's rule-based matching). So it there any documentation on how to use the pattern matcher itself? There's an Odin manual but Odinson seems to have some new pattern grammar rules. Specifically, given a sentence and a pattern, I want to find the tokens that match some sub-pattern.

Thank you!

marcovzla:
Hi!
Thanks for your interest in Odinson. We understand that documentation is important and we are currently working to release some in the very near future. We'll let you know as soon as it's ready.

BeckySharp:
Also @felicitywang in the meantime, if you give me an example of what you're trying to extract maybe I can help you.  It won't be as thorough as good documentation ;) but just something help help until it's ready...?

felicitywang:
@BeckySharp Thank you! So basically, for a pattern and a given sentence, the output I'm looking for is the corresponding strings for each sub-pattern. 

For example(example from README), for pattern
`(?<cause> []) <nsubj [lemma=cause] >dobj (?<increase> [lemma=increase]) >nmod_in (?<theme> []))`, 
and sentence
`Health care progress, antibiotics, vaccination and improving life standards caused a dramatic increase in lifespan in recent decades .`

I'd like the output
```
cause = ["health care progress", "antibiotics", "vaccination", "improving life standards" ]
theme = ["lifespan"]
```

For a more general example, with the pattern `(?<theme>[]) <dobj (?<predicate>[]) >nsubj (?<agent>[])` and any sentence, return a (list of) dictionary {'theme': ..., 'predicate': ..., 'agent': ...}

It seems that this is already done somewhere now that certain tokens are highlighted, yet my lack of knowledge in scala is kinda preventing me from finding/using it. 

Plus, I suppose each dependency arc has only one child, is it easy to find the enclosing phrase the token belongs to(in the example above, `improving life standards` for `standards`)? ([`expansions` in SPIKE-CORD's documentation](https://spike.covid-19.apps.allenai.org/help/search#expansions)).


Suppose I have the above issue resolved, another question is how can I use a different processor, say Spacy?

Thank you.

BeckySharp:
[...]

PR #67, opened 2020-05-08
example entrypoint

BeckySharp:
Allows people to see one way to get started as well as providing a simple way to export results (though that export is an example only and not-comprehensive).

Relates to #65

### DIFF
--- a/extra/src/main/resources/application.conf
+++ b/extra/src/main/resources/application.conf
@@ -14,7 +14,11 @@ odinson.shell {
 
 }
 
+# To run the example, point indexDir to your index
+odinson.indexDir = ${odinson.dataDir}/index
+
 odinson.extra {
+    resources = extra/src/main/resources/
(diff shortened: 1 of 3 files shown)


extra/src/main/resources/application.conf
```diff
@@ -23,4 +27,6 @@ odinson.extra {
     # whether or not to store the Sentence JSON in the index.
     # Though useful developing queries via the webapp, a value of true will greatly increases the size of the index.
     storeSentenceJson = true
+    rulesFile = ${odinson.extra.resources}/example/rules.yml
+    outputFile = example_extractions.jsonl
```
marcovzla, line 31: is `jsonl` a typo?

BeckySharp: no -- it's for json lines

marcovzla: Got it. Thanks!

PR #68, opened 2020-05-08
Changes to REST API to support "event" matcher

myedibleenso:
While functional, there is further work to be done on the REST API.


myedibleenso:
While most of the changes are restricted to the `backend` subproject (the Play app which is home to the REST API), a few changes extend beyond.  For example, note that `IndexDocuments` now retains the name of each json file indexed as part of the document metadata.  This is necessary to retrieve the details for display (parse, token attributes, etc.):

- https://github.com/lum-ai/odinson/blob/5ccdd3cf95372b111ec26f8a02494a1a11474736/extra/src/main/scala/ai/lum/odinson/extra/IndexDocuments.scala#L45-L46

At one point in the past, we stored the json itself.  That approach now retrieves sentence details from the filesystem.

myedibleenso:
Here is a simple Python gist illustrating how one could execute a grammar via a POST:

- https://gist.github.com/myedibleenso/af3456b559455abbb1c931f46beb170a

marcovzla:
Thanks!
LGTM

PR #81, opened 2020-06-13
**testing codcov**

gcgbarbosa:
this is just a test.


gcgbarbosa:
:(

gcgbarbosa:
@marco and @BeckySharp this looks cool.

gcgbarbosa:
@BeckySharp check commit 961cc0a. It has a ‘v’ mark informing that the tests have passed.

BeckySharp:
Cool!  hey -- when I looked at the report, I was able to see how many lines were "hit" and not, but not which lines were hit, is there a way to do that, maybe even with a diff view?

gcgbarbosa:
I don’t know if I understood, but the image I sent over slack has this view of which lines were hit and which were not.

marcovzla:
@gcgbarbosa Can you please add a line to the CHANGES.md file that states what was added with this PR?
Also, is it necessary to change sbt's version?

gcgbarbosa:
> @gcgbarbosa Can you please add a line to the CHANGES.md file that states what was added with this PR?
> Also, is it necessary to change sbt's version?

I am following what they recommend. I don't know if it works with another version.
Should I try keeping the version we were using before?
https://github.com/scoverage/sbt-scoverage

myedibleenso:
@gcgbarbosa, we want to avoid downgrading sbt (that sort of change is in the wrong direction).  Please try testing things with the prior version.

gcgbarbosa:
@myedibleenso, changed it back to 1.2.6. Tests work on my machine. Let's see if it will work with git actions too.

gcgbarbosa:
it worked! @myedibleenso

PR #82, opened 2020-06-16
Start State trait

kwalcock:


kwalcock:
Cool.  I like that the testing is running.  Most of the changes are temporary and I will update their descriptions as they are superseded.

core/src/main/scala/ai/lum/odinson/ExtractorEngine.scala
```diff
@@ -1,10 +1,11 @@
 package ai.lum.odinson
 
 import java.nio.file.Path
+
 import org.apache.lucene.analysis.core.WhitespaceAnalyzer
-import org.apache.lucene.document.{ Document => LuceneDocument }
```
myedibleenso, line 5: @kwalcock, @marcovzla : maybe we should add [a formatting hook](https://scalameta.org/scalafmt/) to avoid an intentional diff war?  I'm of the habit of always adding those spaces, but I think it's more important that we try to be consistent.

kwalcock: I usually put a space there, but would not have changed that on purpose.  I assume that IntelliJ did it without me noticing, which is why I generally do not like tools taking that initiative.  Sorry I didn't notice.  The goal is to adapt to whatever format is present, whether it is wrong or not ;-)  If our tools are conflicting, though, and fighting each other, a tie-breaker might be necessary.  As far as I know, I'm using default settings in IntelliJ and generally haven't noticed many conflicts of this sort.

myedibleenso: Ah, gotcha.  No problem anyway (at least not a pressing one!).

kwalcock:
This message above doesn't look right: marcovzla merged commit d58a6ef into master yesterday.  the commit is not in the list.  Did @marcovzla cherry pick and/or squash the changes or something?  The graph on my display doesn't show the branch reconnecting.

kwalcock:
Usually I'd see a line connecting the branch and master.  (Just want to make sure that everything is syncing up correctly on a new project.)

image

myedibleenso:
@kwalcock, I believe all merges are being squashed.

kwalcock:
If so, I may do it preemptively.  Then I can see that nothing has gone missing.

PR #86, opened 2020-06-18
variables as lists

BeckySharp:
We added this functionality to Odin, because it really helps with diffs etc (and seems like a reasonable representation).
I'd like to port it here, if that's ok.


BeckySharp:
(ignore the `and_lists` of the branch name, I thought I would do something else but decided not to)

BeckySharp:
@marcovzla please check and let me know if this is what you're looking for.
thanks!

PR #87, opened 2020-06-18
Implement a MentionFactory

kwalcock:


kwalcock:
This was quick and a little dirty.  Probably Mention should be a trait.  Eidosdottir can supply its own Mentions, as long as they derive from Mention.  After the change to a trait, they would only have to implement it or derive from an implementation.  @BeckySharp was looking for something like this.  I'm sure the design will evolve as needed.

kwalcock:
That's a cool coverage report BTW.

core/src/main/scala/ai/lum/odinson/Mention.scala
```diff
@@ -39,4 +39,15 @@ case class Mention(
       }
   }
 
+  def copy(mentionFactory: MentionFactory /*= new DefaultMentionFactory*/,
```
BeckySharp, line 42: @kwalcock do I understand correctly that this would allow for making a new "view" of the extraction with a different factory?  If so -- seems like this may be my "entry point" for having a custom mentionfactory/mention in eidosdottir

kwalcock: Probably not, if I understand the question correctly.  This line in ExtractorEngine is where I intended  customization to be incorporated.  The engine gets a single factory when it is constructed which is hopefully good for the engine's entire lifespan.  It might be slipped in elsewhere but this was the main point of attack.

def fromDirectory(config: Config, indexDir: Directory, mentionFactory: MentionFactory = defaultMentionFactory): ExtractorEngine = {

BeckySharp:
ok, cool, thanks!!

core/src/main/scala/ai/lum/odinson/Mention.scala
kwalcock: But that copy method could be used for a one-off.

marcovzla:
is this ready to be merged?

kwalcock:
Yes, if enough people have seen it to conclude that it is a reasonable idea.  It can be further refined.

BeckySharp:
> Yes, if enough people have seen it to conclude that it is a reasonable idea. It can be further refined.

Well I like it :)

marcovzla:
Thanks!

PR #91, opened 2020-06-22
refactoring to simplfy method calls

BeckySharp:
First -- I love the idea of the MentionFactory, I think this will really simply things in external projects!
I am doing this as a PR on your PR branch bc I think you do it that way in Eidos :)

@kwalcock  -- met with @marcovzla and tried to refactor this a little to make it so that Mention doesn't need to care about what MentionFactory you're using when retrieving the arguments.
Did this by (a) making the args a val, not a def, and (b) teaching the MentionFactory trait how to make Mention arguments.  Likely this implementation will suffice most external needs, and won't often need to be overridden, but of course it can be.

If you like this, please merge into your branch and we can then merge into master?

### DIFF
--- a/core/src/main/scala/ai/lum/odinson/DefaultMentionFactory.scala
+++ b/core/src/main/scala/ai/lum/odinson/DefaultMentionFactory.scala
@@ -2,12 +2,18 @@ package ai.lum.odinson
 
 class DefaultMentionFactory extends MentionFactory {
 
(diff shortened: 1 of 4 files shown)


kwalcock:
Thanks for moving things to the right place.  I'm just trying to impress @marcovzla with the code, so if he already knows about it...

PR #92, opened 2020-06-22
renamed the namedcapture

BeckySharp:
Sorry!!!! one more thing, I like these names better!!
(Sorry for the piecemeal, this time it's MY TURN to complain that you're too fast for me!!!)


kwalcock:
capture is a much better name than m, I have to agree.  We can rename the branch so that it looks more shareable, especially if it reemerges from master.

PR #95, opened 2020-06-29
Integrate state

kwalcock:


kwalcock:
Only the bottom three commits are pertinent.  The others are/were part of #82.  They may be repeated here because of the unusual merge back then.

I notice that IntelliJ is reformatting the import statements when it adds a new one so that some spaces around { } are lost.  I'll let that be.

The labels are somewhat unclear.  The rules have them, the extractor has them, and the state has them.  Some of them are the same.  See TestEvent for an example when one was set manually.

marcovzla:
@kwalcock after I merged #94 there is a conflict in this PR.
Can you look at it please?
Thanks!

kwalcock:
Too many cooks in the kitchen or I'm stepping on my own toes, whichever.  Thanks, Marco.

kwalcock:
@marcovzla, aren't the Mentions that are being returned from extractMentions the very things that might be in overabundance and part of the reason that the state is a database rather than a collection of Mentions?  Or is the overabundance at some lower level than that?

kwalcock:
These changes are just baby steps so that @marcovzla can stop me from going too far in the wrong direction.  There are more PRs waiting already.  Don't let the incompleteness prevent a merge.

marcovzla:
> @marcovzla, aren't the Mentions that are being returned from extractMentions the very things that might be in overabundance and part of the reason that the state is a database rather than a collection of Mentions? Or is the overabundance at some lower level than that?

I am not sure I understand the question. Can you give me some context? or we can do it on a call.

kwalcock:
I will delete this PR shortly as being superfluous.

PR #105, opened 2020-07-01
Offline docs

gcgbarbosa:
- Add a `Gemfile` for offline deployment
- Add a doc page describing how to deploy docs offlihne
- Ignore docs temp files


gcgbarbosa:
Nice. 

Btw, this is how I choose colors in my project:
https://material.io/design/color/the-color-system.html#color-usage-and-palettes.

When I am extra lazy I just steal them from here: https://color.adobe.com/explore

Maybe you already know these but just in case...

PR #108, opened 2020-07-03
Loop through priorities

kwalcock:


kwalcock:
@marcovzla, it sounds like you would like some of the state operations to be moved from OdinsonIndexSearcher.odinSearch to its caller in ExtractorEngine.  That sounds good and I can give that a try, and so can anyone else.  I tend to take smaller steps and this step was about looping through the priorities.  The next step was going to be about the event arguments, but I'll the rearrangement of state in between.  I think the smaller steps help with merging and hopefully makes your life easier by not overwhelming with unrelated changes, although the incompleteness may be just as troublesome.

marcovzla:
sounds good, but it seems like this PR overlaps a lot with the other ones.
What are your thoughts on the order of merging?

kwalcock:
Time out.  Let me send you a screen shot.

kwalcock:
@marcovzla, there's something that needs to be sorted out, but I'm not sure what yet.

In general it works better to go from the low numbers to the high ones.  In this case, the PRs are from straight line development.  (I just didn't know how far you wanted me to go.)  In the graph, see that #112 should incorporate all the changes from #108 and #95.  What I don't understand is why I don't see a line between #112 and origin's master (at the top).  That was the case last time as well.  This picture came from GitKraken and I didn't get it 100% integrated with the odinson repo IIRC because of some permissions things.  Maybe that's the problem and I will try to check.  It could be that something else is happening though?  I can't tell what all got merged in the usual way.  Did you just press the merge button at GitHub or was more involved?

image

kwalcock:
I checked the integration and it seems to have been solved previously.  The reason for the lack of a line is that commit cb873fc82108e2164760b62f7397cb601033b967 comes a little out of nowhere.  This is apparently how squashed merges show up.  All the right commit messages are there, so I assume all the code is as well.  This makes merges of PRs #112 and #108 unnecessary.

I'm a little skeptical of the squashing.  I think it's part of the reason that there are/will be merge conflicts about #112 and #108.  Git doesn't understand that these have already been included.  If you are interested in a tool, I've been using GitKraken.  As of a year ago it was still the best one I could find and it's free for a public repo.

kwalcock:
I will delete this PR shortly as being superfluous.

PR #111, opened 2020-07-04
Move state management to a better place

kwalcock:
This PR should not (yet) be considered for a merge.  It is intended to provide a space for discussing various design options.


kwalcock:
I'm going to do this over in a cleaner way...

PR #112, opened 2020-07-04
Limit State to ExtractorEngine

kwalcock:


marcovzla:
@kwalcock does this PR replace #95 ?

kwalcock:
This #112 incorporates #95 and #108.  It's a straight line if you want them all, you can skip to the end.

marcovzla:
:shipit:

kwalcock:
@marcovzla, when it says "marcovzla merged commit cb873fc into master 42 minutes ago" and commit cb873fc is not one in the list, what has happened?  It seems like some information is being lost.

PR #116, opened 2020-07-11
Add testing docs

gcgbarbosa:


gcgbarbosa:
Thanks!!! I will update the example with what you suggested on my next iteration over the docs.

PR #122, opened 2020-07-21
Store events in State

kwalcock:


kwalcock:
This large number of commits is because of the previous squashing merge which prevents git from understanding that only the last one has any changes.  It is also the reason for the merge conflicts.  I will try to fix it. :-(

PR #123, opened 2020-07-21
Store events in State

kwalcock:


kwalcock:
Tests passed on the PR after all.  The problem is only on Windows where one test file was incompatible.  There's a separate PR for that.  I did forget to edit the CHANGES file, though.  I'll go back for that when this gets close to being acceptable.

kwalcock:
In the meantime I found Gus' wiki page about state.  I don't think this version is compatible with what he describes.  It can likely be made that way, though, so don't let that hold anything back except maybe the merge.  Other critique is likely still valid.

marcovzla:
Thanks!

PR #145, opened 2020-07-29
Try to debug a test

kwalcock:


kwalcock:
This is not for merging!

kwalcock:
Probably the "becky-gummy-bears" test sentence just doesn't have what the rule is looking for.

BeckySharp:
We don't need a new document, we can change the Person rule to match `[word=Becky]` ?

PR #147, opened 2020-07-31
Make idProvider live for entire State lifespan

kwalcock:


kwalcock:
This is just a one line bug fix in the state, so I'm going to self-merge it and not interrupt anyone's vacation.

PR #148, opened 2020-07-31
Close the State

kwalcock:


kwalcock:
Make the state close properly and drop the table when that happens.  There are no design changes here, so I'm going to merge.

A different version of the state might not drop the table and instead transfer it elsewhere for safekeeping.  Those variations will come later as will also optimizations.

PR #163, opened 2020-08-13
Protect against a double close

kwalcock:


kwalcock:
I'm considering this a trivial change and merging it without consulting Marco.  This is my confession.

PR #165, opened 2020-08-16
Populate the StateQuery

kwalcock:


kwalcock:
@marcovzla, this change is based on what you told me about what was happening behind the scenes.  It needs some tests.  Can I recruit @gcgbarbosa for that?  It would be good to verify that some returned Mention has an odinsonMatch which is a StateMatch.  I see the StateMatch constructor being called, but never yet with named captures populated.

BeckySharp:
> It would be good to verify that some returned Mention has an odinsonMatch which is a StateMatch. I see the StateMatch constructor being called, but never yet with named captures populated.

@kwalcock we would expect for this to happen if we find a match with an Event query (so it has arguments), then that goes into the state, then we match that found Event with another rule?  If so, I can write a test for that easily.  If not, please clarify what expected behavior you're trying to check and I'll try to put a test in.

kwalcock:
Yes, @BeckySharp, that sounds accurate.  If the test suite was already large enough, something would probably have noticed the missing parts, but it seems like it isn't.  One would need to make sure that the Event was found and also that it still has the eventy things with it, like those namedCaptures and maybe that the Mention.odinsonMatch is a StateMatch, although I'm not sure of that.

BeckySharp:
OK, cool, I prob won't get to it today, but will put it on top of the stack

BeckySharp:
ok, pushed a test, please check and see if it's what you were looking for

gcgbarbosa:
Hi guys, sorry for joining this too late. @kwalcock Let me know if there's something else I can do.

kwalcock:
Excellent!  This test indeed fails if I comment out the code that was just added.  You rule the rules!

PR #168, opened 2020-08-19
Get docId and sentId lazily

kwalcock:


kwalcock:
@marcovzla, I think this is more or less what we discussed this morning.  I have been running George's odinson-tests project's TestApp for timing.  Before this change, my computer averaged 19.3 seconds per iteration on the 100 document corpus.  After, it was 16.2 seconds, even with profiling on.  Feel free to rewrite.

marcovzla:
Thanks @kwalcock!
This looks great.

PR #169, opened 2020-08-20
Reuse the idGetter when doc repeats

kwalcock:


kwalcock:
@marcovzla, this is on top of (in addition to) the LazyIdGetter.  If the same doc comes up again, it would be nice to not look it up twice.  There is a little overhead and it might not happen very often, but this is how it might be done.  One goal was not to clutter up that loop in the ExtractorEngine and that seems to have worked.

marcovzla:
It seems like the `MostRecentlyUsed` class relies on the fact that the docs are traversed in order, right?
Maybe we could add a comment explaining that? it took me a while to figure it out

kwalcock:
Yes.  I was probably programming in a vacuum.

kwalcock:
I added some documentation.  If anyone can think of a better name, please do.  Instead of getOrNew it could be previousOrNew, for example.

It often works better to merge in PR order.  Fewer merge conflicts seem to arise that way.

PR #170, opened 2020-08-20
Add a test for IndexDocuments in `extra`

gcgbarbosa:
This solves part of the issue.
The other part is to make it importable as a fixture.


BeckySharp:
please review -- test seems to be failing, thanks!

gcgbarbosa:
That is very weird. It passes on my machine. Even after just cloning it. I will take a look soon.

myedibleenso:
@kwalcock, any idea why we're encountering that State error here?

kwalcock:
I see an extra `Picked up JAVA_TOOL_OPTIONS: -Dfile.encoding=UTF8` and an all tests passed right before as if sbt is starting over again.  When things like this happen, it seems to be the case than an attempt is made to reload an .so, .dylib, or .dll file.  The database may be using one.  The OS doesn't like that.  If you run `sbt` locally and call `test` twice in a row, the same error probably appears.  (It has for me.)  Sometimes the solution is to add lines like `fork := true` or
```
Test / fork := true // Also forces sequential operation
Test / parallelExecution := false // Keeps groups in their order
//Test / testForkedParallel := true // Allow parallel within group?
```
It would take some experimentation and maybe debugging to figure it out.  It may be that the sql library doesn't do it right or needs to check whether the dll is still loaded.

myedibleenso:
@BeckySharp , before we merge this, we should decide whether or not we want to postpone upgrading processors until syntactic dependency parsing is back (or remove processor options that don't support it?).

BeckySharp:
yes.... I emailed @MihaiSurdeanu to ask when that will be ready (UD basic is in master now, but IDK when enhanced will be and when it will be released).  I'd like to upgrade soon, but it doesn't strictly need to be now.

BeckySharp:
@myedibleenso -- seems we won't have enhanced until the end of the week, most likely.  Thus, I reverted the processors version to 7.4.4 -- but I didn't add the bioCluProcessor back because we're going to upgrade in the immediate future.
This should allow us to merge to master, though

BeckySharp:
@danebell -- as soon as this is merged, please reindex and let me know if there are issues

PR #172, opened 2020-08-21
docId and sentenceId

marcovzla:
added methods to preserve previous api


BeckySharp:
should we go ahead and merge this?

kwalcock:
It looks like I had approved but not pressed the merge button.  Usually I defer to @marcovzla for that, but since he asked...

PR #173, opened 2020-08-22
validating docs before indexing and made test resources smaller

### LINKED ISSUE #156: Preprocessing steps for making an index
As a index-user I want a pipeline that handles some of the known issues with processors, so that I can not have weird issues when using the index.

Example -- some documents end up with empty tokens, which cause off-by-one errors when retrieving the odinson result strings.  This is currently happening with the arapidopsis index:

(note that this query works in other docs, but here see how it thinks that AND is the verb, not phosphorylate)
image

I think that @robertvacareanu ran into this already, and it got fixed in processors, but likely wasn't in the version that Gus ran -- but we preprocess stuff in the Odinson pipeline

So we should do some preprocessing/sanitation

BeckySharp:
This PR addresses several things:
- @myedibleenso 's request for the test resources to be smaller (I made smaller docs with slices of the sentences) and with splitted => split
- the text is now validated, such that when a string is composed of only whitespace (and invisible control characters) we replace it with another character (set in the config).  This is used to fix the off-by-one error caused by lucene omitting these tokens while they still existed in the odinson documents.  By not removing them, only replacing with a diff character, we are still able to use the dependency parse, etc.
- updating to processors 8.1.3 -- this means that we shouldn't have these issues with control chars moving forward, when annotating text, however, this means that BioCluProcessor is no longer available.  If we want it we can add it with a reach dependency I think @myedibleenso 

FYI @gcgbarbosa -- I think this maybe addresses the outstanding todos on your PR #170 

FYI @marcovzla 

closes #156


myedibleenso:
@BeckySharp, I wanted to mention that the version of processors that you switched to likely only supports syntactic dependency parsing in the CoreNLP-derived processors.  I discovered this recently when I upgraded in another project and ended up rolling my own processor as a hybrid of Clu, CoreNLP, and some glue.  That said, I believe the next release of `CluProcessor` will again include some form of dependency parsing (UD but not UD enhanced?).  Just wanted to draw attention to this, as this will likely frustrate users who are unaware of the change.

PR #176, opened 2020-09-01
Docs

BeckySharp:
Added more content to the docs and refaced the main README landing page.
Content from the old one was moved/removed:
- REST API, docker, and shell info all got their own pages in ghpages
- UI info was removed bc I think it's no longer in service?

Added:
- Event queries (including priorities)
- the stuff I moved from README (see above)
- expansion
- a walkthrough example (as requested by @MihaiSurdeanu )
- edited the test page, updating content (as it continually evolves!)

** Bonus: I added a convenience method in the OdinsonIndexWriter to index a File (thanks for the push @myedibleenso and the suggestion about it being a File @marcovzla !), which cleaned up the IndexDocuments app a bit and should help out what @maxaalexeeva is working on...

Please check out the readme at least and let me know if it's ok, the docs obviously aren't done, but baby steps, right?
It will be nice to have the info about Event Queries, expansion, etc in the docs.


maxaalexeeva:
@BeckySharp, you mention that UI info has been deleted because it is no longer in service. Is that the sample docker commands for annotating and indexing that are no longer there?

myedibleenso:
maxaalexeeva , no, we used to include a UI, but we stopped maintaining it.

maxaalexeeva:
ah! Should I expect this to still work, @myedibleenso ? 

```bash
docker run \
  --name="odinson-extras" \
  -it \
  --rm \
  -e "HOME=/app" \
  -v "/path/to/data/odinson:/app/data/odinson" \
  --entrypoint "bin/index-documents" \
  "lumai/odinson-extras:latest"
```

myedibleenso:
>ah! Should I expect this to still work, @myedibleenso ?

If you build the `lumai/odinson-extras:latest` docker image with `sbt dockerize` (as described on [this page](http://gh.lum.ai/odinson/docker.html)), yes.  

We haven't published an image for the `extras` subproject, though we do have a repository on DockerHub.

maxaalexeeva:
@myedibleenso thanks! I'll try that!

PR #177, opened 2020-09-02
Deprecate old method names before removal

myedibleenso:
Old methods names renamed in 89ceb72095d603cf61d27decc7c42c5eea50c87a have been added back with a deprecation warning (deprecated annotation).

As described in #174 , these were renamed in 89ceb72095d603cf61d27decc7c42c5eea50c87a to distinguish between overloaded versions of methods of the same name that return **all** tokens or characters in a sentence vs these two which only return the tokens or characters within the span of some mention/match.


marcovzla (approved):
Thanks!

PR #182, opened 2020-09-04
Add state support for Argument Promotion

kwalcock:


kwalcock:
It looks like I missed one :-(  It's wasn't directly related to state storage, so would have happened regardless of mechanism.  Good catch and good test.

BeckySharp:
@kwalcock thanks so much for figuring this out so quickly!
Once I saw your fix, I wanted to check the fullTraversal too.  I added one more test, it failed, and then I added that `setState` as well.  Again, thank you so much for figuring this out!!
@marcovzla I think this is ready for you to review!

kwalcock:
Or two.  Thanks!

BeckySharp:
hey @kwalcock -- I know how much you like failing tests and PRs that linger... so.
I found that some event arg semantics that were designed but not fully implemented, were .... not fully implemented :)
Basically, we intended to be able to "promote" arguments in events, meaning that we could find them in the moment, tell the Engine what their label was, and they would be added to the State in their own right, alongside the events they participate in.

I added a failing test to illustrate the desired behavior -- note that the `^` syntax is what signals the promotion of the argument.

To make this happen, @marcovzla  and I just now passed this "flag" through to the ArgumentSpans, and then to the ArgumentMetadata that each EventMatch has.  That means that now, in the ExtractorEngine, wherever you have matches, if you check to see if they are EventMatches you can treat them special.

We left you a verbose comment in ExtractorEngine, in a `query` method.... can you please help?

THANKS!

kwalcock:
It has been added to the list.  Can I trade it for new groundings ;-)

BeckySharp:
Yes! Updates forthcoming today

kwalcock:
I would merge this branch just after the "minutiae" commit.  This seems like a very different issue.

kwalcock:
I'm not sure this deduplication is feasible at the state level: (i.e., that the arguments weren't already in the state, but we want to add them).

BeckySharp:
At this point we know that they weren't in the state bc we check to see if
it's a StateQuery
also -- if you want to split this off, can you suggest how I would do that
without losing changes?

kwalcock:
I split off at the previous commit and it can be merged via https://github.com/lum-ai/odinson/pull/184.

kwalcock:
Duplicates could have been added to the state at a previous priority or with rules that are copies of each other or just find the same event, or not?  The distinct applies only afterwards to mentions.

BeckySharp:
continuing convo at #184

BeckySharp:
> Duplicates could have been added to the state at a previous priority or with rules that are copies of each other or just find the same event, or not? The distinct applies only afterwards to mentions.

yes -- there may be duplicates in the State, but if you add the arguments that have `promote` in their metadata, if the code does what it was designed to do (we should add a test), _those arguments_ won't be already there, as in they won't be making any existing duplication worse.

The part of the code that "ensures" this (**again**, we should add a test), is here: https://github.com/lum-ai/odinson/blob/53bb6b46c36d4debe9f2b67e277af4446cfb74e6/core/src/main/scala/ai/lum/odinson/OdinsonMatch.scala#L83

where the intended behavior is that if an argument came from the state, then it won't have `promote == true` in the first place

kwalcock:
Verbose comments have been greatly appreciated.  After I added code to pass this new test, I found that others fail.  The assertion that this comment recommends turns out not to hold:
```
    //  Each EventMatch has an Array of ArgumentMetadata (that we suspect is parallel to the Array of named captures
    //  (though we should prob check...), and inside this is an attribute `promote`.  If true, then that argument needs
```
I'm seeing EventMatches with two captures and three metadata and their orders also being scrambled.  For one in EventTest there are captures for "object" and "subject" and then metadata for "subject", "object", "object".  This seems to happen even before I add my code.  I'm adding a branch event_match_bug (https://github.com/lum-ai/odinson/tree/event_match_bug) and an issue (#186).

core/src/main/scala/ai/lum/odinson/lucene/search/MatchSelector.scala
```diff
@@ -143,15 +143,15 @@ object MatchSelector {
       m <- pickMatches(g)
     } yield m
     val packages = arg match {
-      case ArgumentSpans(name, label, min, Some(max), _) if min == max =>
+      case ArgumentSpans(name, label, min, Some(max), _, _) if min == max =>
```
BeckySharp, line 146: nice catch!

core/src/main/scala/ai/lum/odinson/state/EventPromoter.scala
```diff
@@ -0,0 +1,102 @@
...
+    def promoteOdinsonMatch(odinsonMatch: OdinsonMatch): Array[LabeledNamedOdinResults] = {
+      val result = odinsonMatch match {
+        // This is only ever considered in the case of an EventMatch.
+        case eventMatch: EventMatch =>
+          val names = eventMatch.namedCaptures.map(_.name)
+          // It is required that there are no duplicate names in the namedCaptures.
+          // This is not true!
+          // require(names.distinct.size == eventMatch.namedCaptures.length)
+          // Because of this, the namedCapture can be stored more than once in the state.
+          val nameToPromote = eventMatch.argumentMetadata
```
BeckySharp, line 54: @kwalcock -- i haven't fully traced this, so IDK, but this seems to be slightly overly complicated for the task...?
Was there some situation you were trying to handle that led to this, rather than a more straighforward loop with a push to the state?

core/src/main/scala/ai/lum/odinson/lucene/search/MatchSelector.scala
kwalcock: Compilers work wonders.

core/src/main/scala/ai/lum/odinson/state/EventPromoter.scala
kwalcock: See also #186.  Yes, it is overly complicated.

kwalcock:
P.S. I also have to clean up some messy array creation before this would ever be merged.  The point was to illustrate some of the issues first with the PR.

BeckySharp:
@kwalcock please add your awesomeness now that ArgumentMetadata semantics are clarified and enforced. thanks!

PR #183, opened 2020-09-04
implementation of statematch in the match seclector

BeckySharp:
filled in TODO implementation


marcovzla:
I will add a test to check the `EventMatch` case.

PR #187, opened 2020-09-07
Update memory state

kwalcock:
It has also been made the default


kwalcock:
Scrap this.  I'm going to squash ahead of time.

PR #189, opened 2020-09-08
deleted QueryCompiler tests

BeckySharp:
We have thought more and decided that for now, we don't know a good way to test the QueryCompiler. For now we'll remove the tests, and if we come up with a good idea, we can revisit.


marcovzla (approved):
Thanks!

PR #191, opened 2020-09-09
Make TestIndexDocuments pass for Windows

kwalcock:


marcovzla:
Hey @kwalcock. I don't understand how this fixes the issue.
Can you explain it?

kwalcock:
I think that File just does a better job taking the OS into account than Path.  I didn't look into it much further except to verify that it worked.

marcovzla:
What is your `dataDir` value in the config file?

kwalcock:
I have not changed any values in the resource files.  This is the error message:
```
Received dataDir as a parameter </E:/Users/kwa/Clients/lum/Projects/odinson-project/odinson/extra/target/scala-2.12/test-classes/>
[info] - should get the correct list of files *** FAILED *** (427 milliseconds)
[info]   java.nio.file.InvalidPathException: Illegal char <:> at index 2: /E:/Users/kwa/Clients/lum/Projects/odinson-project/odinson/extra/target/scala-2.12/test-classes//index
[info]   at sun.nio.fs.WindowsPathParser.normalize(Unknown Source)
[info]   at sun.nio.fs.WindowsPathParser.parse(Unknown Source)
[info]   at sun.nio.fs.WindowsPathParser.parse(Unknown Source)
[info]   at sun.nio.fs.WindowsPath.parse(Unknown Source)
[info]   at sun.nio.fs.WindowsFileSystem.getPath(Unknown Source)
[info]   at java.nio.file.Paths.get(Unknown Source)
[info]   at ai.lum.odinson.OdinsonIndexWriter$.fromConfig(OdinsonIndexWriter.scala:269)
[info]   at ai.lum.odinson.extra.IndexDocuments$.delayedEndpoint$ai$lum$odinson$extra$IndexDocuments$1(IndexDocuments.scala:40)
[info]   at ai.lum.odinson.extra.IndexDocuments$delayedInit$body.apply(IndexDocuments.scala:16)
[info]   at scala.Function0.apply$mcV$sp(Function0.scala:39)
[info]   ...
```
`  val resourcesFolder = getClass.getResource("/").getFile` returns something with a trailing '/', so somewhere a `ConfigValueFactory.fromAnyRef(resourcesFolder + "/index")` probably has an extra one.

BeckySharp:
I think this website maybe addresses the problem: https://bugs.openjdk.java.net/browse/JDK-8197918

Essentially:
> On Windows systems this call: 
...getResource("/resource.txt").getPath(); 
returns path beginning with "/C:/..". 

> Such path representation is fully accepted by, for example File class but Paths.get is throwing exception. 

> Removing first character "/" is helping but in such case we will get error on *nix systems as we remove root. 

> STEPS TO FOLLOW TO REPRODUCE THE PROBLEM : 
On windows system get any resource path using getResource().getPath() and try to pass it to Paths.get method. 

> EXPECTED VERSUS ACTUAL BEHAVIOR : 
EXPECTED - 
Get correct absolute path to resource. 
ACTUAL - 
java.nio.file.InvalidPathException: Illegal char <:> at index 2: /C:/... 


Note in particular this: 
> Such path representation is fully accepted by, for example File class but Paths.get is throwing exception. 

The responder basically said "this isn't a bug, you should use it differently":

> This is not a bug as " /C:/sample.csv" is not a legal path name, the IllegalPathException is correct. 

> The issue here is that the user's test case is attempting to use a URL path has a file path. The code should be: 
    `Path path to Paths.get(sample.toURI());`


So -- @marcovzla and @kwalcock , IDK what  the action item is, but this hopefully sheds light

kwalcock:
Wow, @BeckySharp, thanks for finding that.  It's reassuring to know it's not just my system.  I think this reinforces that one possible workaround is to go through file instead of directly to path which is more sensitive.
```scala
    val indexPath = config[Path]("indexDir")
```
becomes
```scala
    val indexPath = config[File]("indexDir").toPath
```
On first glance there doesn't seem to be an easy way to slip in the `.toURI()`.

BeckySharp:
should it happen here or, as Marco mentioned earlier, in common?

marcovzla (approved):
I will approve this PR, but I would like to find a fix in [lum-ai/common](https://github.com/lum-ai/common)

PR #193, opened 2020-09-14
clarify expectations of argmetadata

### LINKED ISSUE #186: For EventMatch, namedCaptures.length != argumentMetadata.length
For the branch event_match_bug (https://github.com/lum-ai/odinson/tree/event_match_bug) which comes from state_issue, the argumentMetadata isn't synchronizing properly with the namedCaptures.  Their counts are off and their orders are mixed up.  Set a breakpoint at line 49 of OdinsonMatch to watch it happen.

BeckySharp:
OK, so with these fixes, we shouldn't have the case that there are multiple argMetadatas with the same name because we're:
1. disallowing multiple arguments with the same name and
2. calling `distinct` on the metadatas when we make the EventSketch (previously, before the `MatchSelector` packages the args, when there are multiple ways an argument can be matched (as was the case with the optional `gummy?` in the TestEvents test), there were repeated argument metadatas).

I added two tests for these desired behaviors.  I think now @kwalcock may be able to simplify the promotion in #182 ...?

I think this closes #186

### DIFF
--- a/core/src/main/scala/ai/lum/odinson/OdinsonMatch.scala
+++ b/core/src/main/scala/ai/lum/odinson/OdinsonMatch.scala
@@ -78,7 +78,9 @@ class EventSketch(
   val start: Int = trigger.start
   val end: Int = trigger.end
   val namedCaptures: Array[NamedCapture] = OdinsonMatch.emptyNamedCaptures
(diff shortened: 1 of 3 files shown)


marcovzla (approved):
Thanks!

kwalcock:
What happened to promote?  The distinct will not help because the same name could just have different min or max values.

PR #197, opened 2020-09-15
odinsonexception and a test

### LINKED ISSUE #195: OdinsonException
We should add an `OdinsonException` that extends `RuntimeException` so that when we see errors we know if they're from Odinson or not -- importantly, we would then modify the tests that catch exceptions.

BeckySharp:
IDK if the test was necessary, if it's just cluttering, we can remove.

Also -- if people want this moved somewhere else, just let me know.  I can move it trivially.
Currently in `package ai.lum.odinson.utils.exceptions`

Replaced several RuntimeExceptions in the code and the tests.  There are some other more specialized exceptions which may or may not get replaced.

closes #195

### DIFF
--- a/core/src/main/scala/ai/lum/odinson/RuleReader.scala
+++ b/core/src/main/scala/ai/lum/odinson/RuleReader.scala
@@ -10,6 +10,7 @@ import org.yaml.snakeyaml.Yaml
 import org.yaml.snakeyaml.constructor.Constructor
 import ai.lum.odinson.compiler.QueryCompiler
 import ai.lum.odinson.lucene.search.OdinsonQuery
+import ai.lum.odinson.utils.exceptions.OdinsonException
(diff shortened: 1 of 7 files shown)


marcovzla (approved):
Thanks!

kwalcock:
That rascal marcovzla beat me to it :-)  Those are great tests which I probably would have been too lazy to write.  There is one thing that might be added.  The StateException is going to be rewrapping an underlying SQLException (or something else depending on which state is being used).  There are already some variations implemented: TransactionException and RollbackException and the like.  They need to keep track of the underlying cause.  Could that be added?  I can as well.

```scala
class StateException(message: String, cause: Throwable)
    extends RuntimeException(message, cause)
```

BeckySharp:
I can't find them... are they merged?  you can def have them extend OdinsonException, or if you point me to them I can do I, I'm about to do a small PR for one more Exception I found and some sys.errors

PR #198, opened 2020-09-15
Don't write state in query

kwalcock:


core/src/main/scala/ai/lum/odinson/ExtractorEngine.scala
```diff
@@ -142,10 +142,20 @@ class ExtractorEngine(
     val filter = filters(allowTriggerOverlaps)
     val mruIdGetter = MostRecentlyUsed[Int, LazyIdGetter](LazyIdGetter(indexSearcher, _))
 
+    def localQuery(extractor: Extractor, state: State): OdinResults = {
```
BeckySharp, line 145: if this is where results are added, can we rename to something like `queryAndAddToState` (or something less verbose but still indicating that this is the place where we add things to the state)?
I sometimes have a hard time finding the state-adding, I forget where it is and look in the `extract` method, but it's not there (and that may be because it's a `for... yield` loop), else we could maybe add a comment in the extract loop...?
(Feel **FREE** to ignore if I'm being dumb)

kwalcock: Of course.  Local was a hint to look nearby but it should say what as well as where.

kwalcock:
Don't fear long names.

BeckySharp:
> Don't fear long names.

:) I just talked more at length with @marcovzla about some of the expected behavior of the extraction algorithm, etc, and I think I'm going to try to capture the content of that convo as  PR onto this branch.  that said, once I make that PR, we should talk about it and you should feel MORE than free to mention any and all concerns that I didn't think about :)

more soon!

kwalcock:
Keep in mind that #182 is going to want to add something to that `localQueryAndAddToState`.

```scala
    val labeledNamedOdinResultsSeq = eventPromoter.promoteEvents(labelOpt, nameOpt, odinResults)
    val odinResultsIterator = new SuperOdinResultsIterator(labeledNamedOdinResultsSeq)
```
`extractMentions` was made to parallel the original version.

BeckySharp:
> Keep in mind that #182 is going to want to add something to that `localQueryAndAddToState`.
> 
> ```scala
>     val labeledNamedOdinResultsSeq = eventPromoter.promoteEvents(labelOpt, nameOpt, odinResults)
>     val odinResultsIterator = new SuperOdinResultsIterator(labeledNamedOdinResultsSeq)
> ```
> 
> `extractMentions` was made to parallel the original version.

I think this is fully compatible with what i have, but we can check.  should be able to push it VERY soon!!!!! (so exciting!!!)

PR #199, opened 2020-09-15
Travis CI changes

myedibleenso:
## Summary of changes

- run tests and build docker images using `sbt dockerize`
- publish [`odinson-extras`](https://hub.docker.com/repository/docker/lumai/odinson-extras) and [`odinson-rest-api`](https://hub.docker.com/repository/docker/lumai/odinson-rest-api) docker images to dockerhub when commits are made to `master`

### DIFF
--- a/.github/workflows/scala.yml
+++ b/.github/workflows/scala.yml
@@ -12,10 +12,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
+    - name: Set up JDK 11
       uses: actions/setup-java@v1
       with:
-        java-version: 1.8
+        java-version: "11"
+        ava-package: jdk
     - name: Run tests
       run: sbt coverage test
     - name: Coverage Report
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: scala
 services:
   - docker
 
-#jdk: openjdk11
+jdk: openjdk11
 
 env:
   global:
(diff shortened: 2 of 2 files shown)


myedibleenso:
@marcovzla , are we switching completely to github actions?  If so, do we want to remove or disable travis?

.github/workflows/scala.yml
```diff
@@ -12,10 +12,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
-    - name: Set up JDK 1.8
+    - name: Set up JDK 11
       uses: actions/setup-java@v1
       with:
-        java-version: 1.8
+        java-version: "11"
+        ava-package: jdk
```
BeckySharp, line 19: typo here

BeckySharp:
hey -- the change to java 11 only affects the docker you're building right? bc in extra we still need java 8 (until we can get rid of the corenlp dependency)

myedibleenso:
@BeckySharp , I've used `processors/CoreNLP` with Java 11 and haven't run into any issues (that I've noticed anyway).  What problem(s) did you encounter?

BeckySharp:
my accounts are mostly second hand, through people trying to use tools and having java 11 and then they downgrade to 8 and it works.  @kwalcock @MihaiSurdeanu @marcovzla do any of you remember this?

maxaalexeeva:
That happened to me last week in another project after I updated to java 11 for the docker to run here. An error like this https://github.com/stanfordnlp/CoreNLP/issues/367, so I had to make a separate env for java 8.

myedibleenso:
Interesting.  I haven't encountered that problem yet, though I am using CoreNLP with Java 11 in another project to create a hybrid `Processor` (part CLU and part CoreNLP).  

To expose this issue, I want to try to trigger it by adding annotation tests to `extra`.

kwalcock:
I haven't especially tried Java 11 because of the potential licensing and support issues.

myedibleenso:
@kwalcock , what is the licensing issue?  I think the Classpath exception puts our use in the clear.  As far as support, I use a flavor of OpenJDK.

kwalcock:
That one might need to know what a Classpath exception is or pay attention to which kind of Java is installed are issues themselves.  They aren't major, but I wouldn't switch just because it's there.

kwalcock:
FWIW I'm looking into a problem that reach has with Java 11 and also see this: https://github.com/scala/scala-dev/issues/559.

PR #202, opened 2020-09-16
Rearrange the extractMentions for state/noState

BeckySharp:
@kwalcock As promised, sir, here is something that I think does two things desired:
- uses iterators instead of Seq so we don't run out of memory
- IMPORTANT, doesn't add the results to the state until _after_ the entire epoch/priority is over, so that extractors from the same priority don't have (inappropriate) access to things found contemporaneously (what a big word, did I use it right?) :) 

1. Please look it over, scrutinize details and all...
2. I added a method to the State trait (`def getAllResultItems(): Iterator[ResultItem]`), I have an implementation in MockState (I'm just that awesome!), and one in MemoryState (that you should check), BUT I left the SQL one for you so you could have fun too.  You're welcome! :)

Thanks, and let's keep in touch!

### DIFF
--- a/backend/app/controllers/OdinsonController.scala
+++ b/backend/app/controllers/OdinsonController.scala
@@ -215,7 +215,11 @@ class OdinsonController @Inject() (system: ActorSystem, cc: ControllerComponents
         case None => extractorEngine.numDocs()
       }
 
-      val mentions: Seq[Mention] = extractorEngine.extractMentions(composedExtractors, numSentences = maxSentences, allowTriggerOverlaps = allowTriggerOverlaps, disableMatchSelector = false)
+      val mentions: Seq[Mention] = {
+        // FIXME: should deal in iterators to allow for, e.g., pagination...?
+        val iterator = extractorEngine.extractMentions(composedExtractors, numSentences = maxSentences, allowTriggerOverlaps = allowTriggerOverlaps, disableMatchSelector = false)
+        iterator.toVector
+      }
       
       val duration = (System.currentTimeMillis() - start) / 1000f // duration in seconds
 
--- a/core/src/main/scala/ai/lum/odinson/ExtractorEngine.scala
+++ b/core/src/main/scala/ai/lum/odinson/ExtractorEngine.scala
@@ -17,9 +17,10 @@ import ai.lum.odinson.compiler.QueryCompiler
 import ai.lum.odinson.lucene._
 import ai.lum.odinson.lucene.analysis.TokenStreamUtils
 import ai.lum.odinson.lucene.search._
-import ai.lum.odinson.state.{MockState, OdinResultsIterator, State, StateFactory}
+import ai.lum.odinson.state.{MockState, MockStateFactory, OdinResultsIterator, ResultItem, State, StateFactory}
 import ai.lum.odinson.digraph.Vocabulary
 import ai.lum.odinson.utils.MostRecentlyUsed
+import ai.lum.odinson.utils.exceptions.OdinsonException
 
 
 class LazyIdGetter(indexSearcher: OdinsonIndexSearcher, documentId: Int) extends IdGetter {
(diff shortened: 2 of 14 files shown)


kwalcock:
The branch name is unacceptably long.  Please rename!  :-)))

kwalcock:
The MemoryState implementation is going to conflict with #200, which is unfortunate.  I would need to write a test for the getAllResultItems() to make sure and it seems like we'll need one somewhere before merging, maybe after the SQL gets one.

kwalcock:
I think the MockState idea is/was brilliant and I would use it even more.  Right now it looks like it is making the code more complicated rather than less.  It fits the null object pattern very well.  I suggest having the ExtractorEngine object keep a MockStateFactory instance around for use when there is supposedly no state.  So, if possible, there wouldn't be an extractWithState and extractNoState distinction.
```scala
  def extractMentions(extractors: Seq[Extractor], numSentences: Int, allowTriggerOverlaps: Boolean, disableMatchSelector: Boolean): Iterator[Mention] = {
    val tmpStateFactory =
        if (needsState(extractors)) stateFactory
        else ExtractorEngine.mockStateFactory

    extractWithStateFactory(extractors, numSentences, allowTriggerOverlaps, disableMatchSelector, stateFactory)
  }
```
However, maybe you already thought of this and it wasn't going to be beneficial.

kwalcock:
That `needsState` is an excellent idea.

kwalcock:
In `extractWithState`, the iterator that state.getAllResultItems() returns cannot be allowed to exist after the state is closed.  The database connection will be closed, the tables deleted, the files erased, etc. by then.  Really I should wipe the memory state as well for consistency.  Uh oh.  `mkMentionsAndFilter` also returns an iterator.  This may not work.  I had things like
```scala
def getDocIds(docBase: Int, label: String): Array[Int] // TODO: Return iterator
```
because I hadn't figured out a solution.  The best strategy I found was to allow the state to be used by an external function, which is the reason for
```scala
  def usingState[T](function: State => T): T
```
Is there any chance that such a function can be passed in to any of the extract methods?  "I'll give you an iterator, but you have to tell me in advance what you are going to do with it, because I'm going to close as soon as you are done, like it or not".

kwalcock:
It's cool how iterators can be concatenated!

BeckySharp:
> I think the MockState idea is/was brilliant and I would use it even more. Right now it looks like it is making the code more complicated rather than less. It fits the null object pattern very well. I suggest having the ExtractorEngine object keep a MockStateFactory instance around for use when there is supposedly no state. So, if possible, there wouldn't be an extractWithState and extractNoState distinction.
> 
> ```scala
>   def extractMentions(extractors: Seq[Extractor], numSentences: Int, allowTriggerOverlaps: Boolean, disableMatchSelector: Boolean): Iterator[Mention] = {
>     val tmpStateFactory =
>         if (needsState(extractors)) stateFactory
>         else ExtractorEngine.mockStateFactory
> 
>     extractWithStateFactory(extractors, numSentences, allowTriggerOverlaps, disableMatchSelector, stateFactory)
>   }
> ```
> 
> However, maybe you already thought of this and it wasn't going to be beneficial.

I can't take credit for the MockState -- that was all @marcovzla 
the reason there's still the state/no state is the return of the iterators, if there's a mock state, and I ask it to give me all the mentions, it's a problem.  Also, with the noState we don't have to loop through anything.  That said, we need to come up with a solution for the iterators anyway...

BeckySharp:
could we close the state a different way?
like, could we make a custom iterator that closes the state when empty?
sorry... just trying to think outside the box here

kwalcock:
Ah, my bad.  All results are going through the state now.  It's not that they are being tee'd.  So MemoryState is a possibility, but that might indeed be inefficient.  There could eventually be a different kind of shim that helps smooth the code.

kwalcock:
What do we do with the labelled, promoted event parts that you don't see here yet?  They were needing to be added to state but are there only for being found by other rules and should not be returned with the rest of the results are far as I know.  Do they need to be flagged in the database as being exceptions.  For MemoryState they might need to be stored separately or maybe just filtered.

BeckySharp:
> What do we do with the labelled, promoted event parts that you don't see here yet? They were needing to be added to state but are there only for being found by other rules and should not be returned with the rest of the results are far as I know. Do they need to be flagged in the database as being exceptions. For MemoryState they might need to be stored separately or maybe just filtered.

They should be returned with the rest of the mentions.  In the future, we can add something like the `keep: Boolean` in Odin Mentions, but I don't think we need it yet.  But, the intended semantics of "promote" is that they will go into the State.  We should clarify that in the documentation and in some scala docs

BeckySharp:
> could we close the state a different way?
> like, could we make a custom iterator that closes the state when empty?
> sorry... just trying to think outside the box here

something like this:

```scala
class OdinResultsIterator(labelOpt: Option[String], nameOpt: Option[String], odinResults: OdinResults, closes: Option[State]) extends Iterator[ResultItem] {

  override def hasNext: Boolean = {
    val hasNext = 0 < matchesRemaining
    if (hasNext) return hasNext
    
    if (closes.isDefined) {
      closes.get.close()
    }
    
    hasNext
  }

```

@kwalcock @marcovzla

kwalcock:
I think it would be frowned upon.  There is no guarantee that anyone will iterate through the whole thing.  They could just .take(5).  If they will, they could just change
```scala
val resultItemsIterator = state.getAllResultItems()
resultItemsIterator.foreach { resultItem => f(resultItem) }
```
to
```scala
state.withResultItemsIterator { resultItemsIterator => resultItemsIterator.foreach { resultItem => f(resultItem) }}
```
but then with a few extra _s.

Do we have a lot of example uses already?  You are probably user #1.

BeckySharp:
I mean, if they don't take all the elements, we need the state to stay
open, right?
Like, if we do something with pagination, while they're consuming we need
to leave the state open?
And when they kill the program, even if the iterator is not fully consumed,
it would close then, right?

BeckySharp:
> Do we have a lot of example uses already?

we have some, yes, but IDK what you'd like me to detail from them.  An issue that has come up (with the taxero use case?) is that with big queries we're running out of memory, which I think was tied to the non-iterator nature of the `extractMentions` method  (unless I'm remembering wrong).

kwalcock:
I think we should have a conference call with Marco.  The comments above imply that something else is managing the lifetime of the state.  The StateFactory should be altered to return a state and then the ExtractorEngine would need to be able to use this existing state.  One can't have the usingState at the same time as the iterator which outlives the state.

Just because I can't delete anything, here's the previous version...

Marco has an opinion, I'm sure.  I don't know the use cases.  No, the state doesn't need to stay open if someone doesn't take all the elements.  Can the pagination stay within a function?  That sounds difficult and it is the reason some of the other state methods return arrays.  Things would close, but not cleanly.  There would be database tables left on the drive, for instance.  Can you keep track of open iterators on a per connection basis and close them?  It sounds like a stateful connection is implied.  Perhaps we have to paginate out of the state.  The request for the iterator or the array includes a page...

marcovzla:
Hi @kwalcock 
We started a document to work through the various possibilities for the state, hitting on what you said above about use-cases.
I think @BeckySharp mentioned that she can incorporate these into a draft for us to discuss.
We can try to talk about it tomorrow. Any time after 1pm would be good.
This is the link to the doc: https://github.com/lum-ai/odinson-development/wiki/State
Please take a look and feel free to contribute.

kwalcock:
I hesitate to edit a wiki page and wish a little that it was a google doc where comments could be added and revisions suggested.  It does seem like something that needs a little more top down design (e.g., the discussion you suggest) than bottom up.

BeckySharp:
well, the wiki is private, and also you can feel free to add comments!!! :)

PR #203, opened 2020-09-16
replaced sys.error and silenced warning

BeckySharp:
FYI @kwalcock , I couldn't find the Exceptions you mentioned in #197


kwalcock:
They are still in a branch, but they will need the cause argument in order to inherit from OdinsonException.

BeckySharp:
@kwalcock -- does this work for your needs in #197 ?
SO was my bff here, feel free to edit!

kwalcock:
Thank you!  It has been so long since I've seen a null.  It seems like a relic from the past.

kwalcock (approved):

marcovzla (approved):
Thanks!

PR #205, opened 2020-09-20
Save state

BeckySharp:
As with previous PRs and the discussion, there the engine adds the mentions to its state, the State lasts the life of the Engine, and optionally can be persisted or dumped.
- Not all methods are in place (e.g., some SQLState methods and the jsonlines serialization of Mentions)
- I'd like someone to check/help write tests for the Ordering adaptation for Mentions vs ResultsItems in MemoryState (for the TreeSet)
- I removed the StateFactory (didn't seem necessary)
- I put the ResultsIterator in the MentionFactory, since the Iterator creation is where the Mentions get built, so the MentionFactory seemed like a good place to do it
- I passed the IndexSearcher to SQLState.  This is because the LazyIdGetter needs access to the IndexSearcher, and that's needed to make Mentions, and the SQLState essentially re-makes mentions when asked to "getAll/Some"

Please review and we can discuss!! :)


BeckySharp:
for the sake of a single state branch, merging

kwalcock:
These squashing merges aren't working for me at all.  Based on the commit messages alone, this is where the code has ended up and where new code should be added.  This branch also very much needs a different name.

kwalcock:
I can rename if this is the place to do it.  How about stateV2?

BeckySharp:
yes, that's good.  I think this branch is going to be the "working branch"
for the state

PR #208, opened 2020-09-24
Persist SQL state

kwalcock:


BeckySharp (approved):
Looks awesome!!! Thanks for filling this in!
I'm SOOOO excited to see all this come together!
But, any idea why we're having to ignore tests?

kwalcock:
They seem to have been broken already and were going to be fixed around the same time as getAllMentions() is implemented, which is soon.

`Iterator[Mention]` may need to be a `CloseableIterator[Mention]`.

BeckySharp:
cool, thanks!

PR #210, opened 2020-09-28
State v2

BeckySharp:
opening for "sharing"/facilitation


kwalcock:
@BeckySharp, @marcovzla, with the new design, it is not possible to do something like this from TestEventTriggers.scala:
```scala
    val mentions = ee.extractMentions(extractors).toArray
    val triggers = mentions.map(m => ee.getStringForSpan(m.luceneDocId, m.odinsonMatch.asInstanceOf[EventMatch].trigger))
    val animals = mentions.map(m => ee.getStringForSpan(m.luceneDocId, m.arguments("result").head.odinsonMatch))
```
The odinsonMatches are coming from the state and that only offers a StateMatch while they previously could be any kind of OdinsonMatch and a match/case statement could be used to distinguish them (which is perhaps the reason for the sealed trait).  The original kind of match is not stored in the state and the MentionFactory would not be able to use that information to make some kind of fancy Mention.  This might need to be changed, depending on your plans.

PR #214, opened 2020-10-06
Make State work with iterators

kwalcock:


kwalcock:
I need to write some tests that show what happens when the database is being accessed twice.  I'm using two connections, but I'm not certain that is necessary.  This seems to be working even if it might be somewhat wasteful.

PR #217, opened 2020-10-08
new tests + adjustments to existing tests and related methods

maxaalexeeva:
New tests:
- /backend builldinfo
- /extra annotate text

Other adjustments:
- added passing an argument to AnnotateText (to allow passing a test config)
- added a test document to backend resources
- added a test document to extra resources 
- disabled controller provider module 
- using temp directories for tests
- allowing the use of extras during backend testing
- paths are made without using "+"
- deleted odinson contexts


myedibleenso (approved):
Thanks, @maxaalexeeva !

PR #218, opened 2020-10-15
Promotion + matches from State are StateMatches

BeckySharp:
This is an attempt at a simple "for now" solution to argument promotion into the state that is compatible with the current version of the State (i.e., trying to handle merge conflicts).

Code from #182 included that makes the matches coming out of the state all be `StateMatch`es and that fixes some bugs in the passing of the promotion params.

Undecided are:
- do we expect promotion when we call extractNoState ? meaning, do we expect to be given the args in the results too? or only with the mentions?
- do we expect promotion when we call query

Currently, we don't promote in either situation, but that can be discussed for sure.

Likely this is only semi-compatible with the SQL state, but we can address that in many ways (esp if it's moving to a plugin)


BeckySharp:
per discussion `extractNoState` should still promote but `query` not

PR #228, opened 2020-12-09
Remove MentionFactory, rename MentionsIterator

BeckySharp:
The MentionFactory was intended to allow users to define their own mentions, but was determined to not be as viable a solution as initially conceived.  This PR removes it to simplify the code, especially to not require users to have a MentionFactory around when making copies of mentions, etc.

Additionally:
- moves all IdGetter stuff to one place
- renames and consolidates OdinMentionsIterator to MentionsIterator


kwalcock:
I would have thought that the DefaultMentionFactory could have handled most cases while still leaving room for customization.  The Odin/Eidos Mention divide has been a pain, but maybe you can unite them in this new round.

BeckySharp:
I think that the complications were related to States having to support extra stuff and it's more complicated than we want now

marcovzla (approved):

PR #231, opened 2020-12-16
Optionally store all fields, and change namespace

BeckySharp:
added key to config `odinson.storeAllFields`, that when set to `true` will cause the IndexWriter to store all the sentence fields.  This enable you to quickly access various fields for Mentions/Matches, which enables efficient post-processing (e.g., filtering based on content).

Added some additional methods to `ExtractorEngine` to get those tokens.


marcovzla (approved):
Thanks!

PR #232, opened 2020-12-19
Test library that can be used in external projects

BeckySharp:
Moved test utils to `utils` package in the main project so that other users who use Odinson can use the test utils more readily/intuitively (without jumping through hoops).

As the intent is for external user ease/usage:
- added docstrings to methods
- added a couple of extra helper utils that should make it easier to test the text content of event mentions
- renamed some methods to be more explicit

@marcovzla I can add similar methods for ~~non-event mentions~~ testing mention content differently (labels, no trigger/mention span, etc)  if you think it would be useful, or else we can add things as we use these utils in our IE systems, based on the needs we find ourselves with


BeckySharp:
I guess in an upcoming PR we should write tests for these test utils, which would be in `test` package...
or should I write them before merging while we're thinking about it?

BeckySharp:
@marcovzla added tests for OdinsonTest, and good thing I did because I wasn't properly handling some edge cases.  Made the appropriate fixes

marcovzla (approved):
Thanks!

PR #236, opened 2021-01-11
update documentation and add test for doc walkthough

BeckySharp:
The documentation walkthrough had an issue with the name of the extractor engine, (`ee` vs `extractorEngine`).  I also removed a check to print triggers for EventMatches to simplify things.  

I added a test, this was to check on #235 -- the test passed, will follow up there.


marcovzla (approved):
Thanks!

PR #243, opened 2021-01-26
Token frequency API route for plots

danebell:
This pull request enables a new API call, `termFreq`, which can return counts, percentage-of-corpus, or log-transformed counts for any token field in the index. You can now retrieve the available fields through `corpusInfo` to make this less error-prone. You can also specify a min and max rank for the request, and order by alphanumeric order or by frequency (with an option to reverse).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
+- Added REST API endpoint for returning frequencies of token-based annotations in a corpus.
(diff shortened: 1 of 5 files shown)


myedibleenso (changes requested):
- [ ] We need a test or two for the new endpoint (see the existing tests for examples)
- [ ] Updates the changelog with a note about the new endpoint

Finally, are those formatting changes to the controller really necessary?

danebell:
Thanks, Gus. I didn't intend those formatting changes. I must have auto-indented the whole file instead of the part of the code I was editing.

myedibleenso:
I figured.  I think we should add [Scalafmt](https://scalameta.org/scalafmt/) as a plugin, decide on a config, and add the formatting as an CI hook to enforce a consistent style across the repo.

PR #252, opened 2021-02-11
Linting via scalafmt

myedibleenso:
## Summary of changes
- Added .scalafmt.conf for linting
- Added lint check in CI as github action

## TODO
- [x] Add contributor guidelines referencing use of `sbt scalafmtAll` and/or [format-on-save workflow](https://scalameta.org/scalafmt/docs/installation.html#format-on-save)
- [x] Reach consensus on format config? (we can revisit this later)


kwalcock:
Thanks for putting so much effort into contributing.md.  The one thing that caught my attention was the email address.  I know we don't have an infinite number unless they can be redirected to a catchall address.  dev@ seems to be only indirectly related, enforcement@ is too negative.  How about conduct@ or community@?  They are relatively positive and express more of a first-class concern rather than an afterthought or distraction from normal business.

kwalcock:
For this particular project it would be good to make sure that those who wrote the bulk of the code are satisfied with the formatting options, which means not me.  I was unexpectedly not displeased with the changes, but there were at least two cases where I thought it would have been worth turning the formatting off so that the original was not changed.  On the initial format that would mean doing a diff to look for such things, reverting, protecting the section, and reformating.  It may not be worth it, but here are the two examples of before
```scala
  val docsDir            = config[File]("odinson.docsDir")
  val DOC_ID_FIELD       = config[String]("odinson.index.documentIdField")
  val SENTENCE_ID_FIELD  = config[String]("odinson.index.sentenceIdField")
  val WORD_TOKEN_FIELD   = config[String]("odinson.displayField")
  val pageSize           = config[Int]("odinson.pageSize")

        case (None,      maxOption, "}")  => Pass(GreedyQuantifier(0, maxOption))
        case (Some(min), maxOption, "}")  => Pass(GreedyQuantifier(min, maxOption))
        case (None,      maxOption, "}?") => Pass(LazyQuantifier(0, maxOption))
        case (Some(min), maxOption, "}?") => Pass(LazyQuantifier(min, maxOption))
```
and after

```scala
  val docsDir = config[File]("odinson.docsDir")
  val DOC_ID_FIELD = config[String]("odinson.index.documentIdField")
  val SENTENCE_ID_FIELD = config[String]("odinson.index.sentenceIdField")
  val WORD_TOKEN_FIELD = config[String]("odinson.displayField")
  val pageSize = config[Int]("odinson.pageSize")

        case (None, maxOption, "}")                 => Pass(GreedyQuantifier(0, maxOption))
        case (Some(min), maxOption, "}") =>
          Pass(GreedyQuantifier(min, maxOption))
        case (None, maxOption, "}?") => Pass(LazyQuantifier(0, maxOption))
        case (Some(min), maxOption, "}?") =>
          Pass(LazyQuantifier(min, maxOption))
```

kwalcock:
Thanks also for not just integrating with GitHub but also with CI.  We have multiple systems and it might be a challenge to get all the different ones to work.  I'm imagining letting this one run for a few weeks or months before others are assimilated.  It may be that we don't like it after all or we run into unexpected side effects, etc.  For example, I'm about to see how well this works with Windows, sinister LOL.

kwalcock:
It seems to be working.  You may or may not want to merge PR #253 onto this branch.  Can the others be sold on the idea and the format?

myedibleenso:
No problem.  My thinking is we can further tweak formatting later, but @marcovzla @BeckySharp @danebell feel free to express your opinions.  I basically went with a style I generally liked without going crazy trying to tackle every little thing.

@kwalcock, feel free to try merging this branch into the other one you referenced.

What other CI is involved here, @kwalcock ?  I think github is sufficient for the time being.  

As far as the email address goes, I agree something like community@ would be a better choice.  If there aren't any objections, I'd like to make that change in a separate PR after I've had a chance to create an address alias.

We should probably document the magic comment that insulates code blocks from formatting.  Can you add that to the dev docs, @kwalcock ?

kwalcock:
The intention was to merge a little patch #253 into this one.  (Am I doing it backwards?)  I should check more thoroughly for sections to unformat, though, if examples like those above are OK.

Other projects are using Travis and Jenkins and this would be integrated differently, I presume.

I can add a note to the dev docs.  One of the dependencies of those dev docs is causing a security warning. so I may as well update their versions as well.  It was already done for processors.

kwalcock:
If there was only one more thing I could change in the formatting, it would be to back off on the aggressiveness of line splitting, perhaps raising it to 100, even given the discussion at https://scalameta.org/scalafmt/docs/configuration.html#maxcolumn that argues for the 80.  Perhaps we just have longer, better variable and method names and wider screens.

myedibleenso:
Sure.  I wondered about the same thing.  Please feel free to make that change on this branch.

kwalcock:
How did you go about formatting all documents?  Did it involve temporary changes to build.sbt?  If you are still tooled for that, I/Lum should take advantage of it.

You and George are the only ones who seem to use git submodules.  Normally I would request that you merge my suggested revisions into your branch and let you or Marco merge your branch into master.  I'm wary of attempting that with a submodule.  I don't quite trust myself or GitKraken with them yet.  If submodules are the way to go, could you clue in the rest of us?

myedibleenso:
>How did you go about formatting all documents?

You run an sbt task that I've documented in the [`**Pull Request Process** section of `contributing.md`](https://github.com/lum-ai/odinson/pull/252/files#diff-419d79ae7d440d49b3fa5fa7739f5f3a359e3ca42c3fa942e5604ae3474eb35bR26).

There are two new tasks worth remembering.

1. Format everything according to the config:

```scala
sbt scalafmtAll
```

2. Verify that everything has been formatted according to the config:

```scala
sbt scalafmtCheckAll
```

>You and George are the only ones who seem to use git submodules.

There's no git submodule here.  I think you might be referencing the CI config for GitHub Actions which must be stored under `.github/workflows`.

kwalcock:
Apparently I need to read the instructions three times in order to remember them.  Thanks for the reminder.

The use of a / in the branch name results in it looking a lot like a submodule for me.  However, it also looks very handy for grouping a person's branches.  I will probably start using a slash instead of dash.  Cool.

kwalcock:
If we want to go with 100, it is probably better to go all the way back to just before your "First autolint" commit and run it again.  Formatting to 80 and then to 100 is different than formatting directly to 100.  scalafmt does not try too hard to find lines that it could have written without a break and put them together, so there are still lots of ill effects from the 80.  All the subsequent changes would need to be applied afterwards.  We're meeting tomorrow a few times anyway, so I'd like to hear what others think of these developments before going ahead.

myedibleenso:
@kwalcock , it's now formatted using a max of 100.  Rather than going back in time and replaying commits, I simply checked out each subproject from `master` before rerunning formatting.  When/if this is merged, we'll squash things into a single clean-ish commit.

myedibleenso:
... and I've preserved the formatting you pointed out in `QueryParser`

kwalcock:
Thanks.  There are a couple of other files that were to be touched up in #253 as well as the documentation changes you requested.  On the other hand, it looks like a bot fixed the nokogiri problem and I can now throw out that commit.  So, I'll eventually update that PR accordingly.  Perhaps the general formatting issue can be discussed at that meeting about the release today and then at least your changes can be merged and then built upon.

kwalcock:
If @myedibleenso is interested in getting all of this formatting into a single PR, then #253 has pertinent commits in

https://github.com/lum-ai/odinson/pull/253/commits/da1127a343c68a382fd1da3808c53bfed3fdb9ba - formatting of OdinsonController

https://github.com/lum-ai/odinson/pull/253/commits/bff2e4c0affe5c47d2c7671ce48c801b50ce27ab - formatting of QueryParser

https://github.com/lum-ai/odinson/pull/253/commits/59a7b0ee2e0a2ac229bbb4461664f12df9c2adbf - formatting of ProcessorsUtils

https://github.com/lum-ai/odinson/pull/253/commits/8ef35e2ea92a6b7ea0c27ae0bfd6c2010225611f - documentation of formatting tips

https://github.com/lum-ai/odinson/pull/253/commits/6a773f25ecb7d102c1be8c57494697ec47173f75 - spellcheck of documentation

These could be cherrypicked.  After the last reformatting, they might have merge conflicts.  It would be great if at least the last two were not lost.

myedibleenso:
...and I think that takes care of it!

PR #262, opened 2021-03-01
Plot api

danebell:
There are two changes here:
1. Improvements to the `term-freq` endpoint: additional arguments provide (a) a second, conditioning variable to allow two-way counting, and (b) a regular expression string to filter the terms returned.
2. A new endpoint, `term-hist` which returns a frequency table of terms binned either by user specification or by a sensible default dependent on whether the bins should be fixed-width or proportional.

Tests are included.


backend/app/controllers/OdinsonController.scala
```diff
@@ -93,6 +97,20 @@ class OdinsonController @Inject() (config: Config = ConfigFactory.load(), cc: Co
     pretty: Option[Boolean]
   ) = Action.async {
     Future {
+
+      // if a regex filter is specified, apply it
+      def isMatch(s: String, filter: Option[String]): Boolean = {
```
myedibleenso, line 102: Let's add a docstring to this method.

backend/test/controllers/OdinsonControllerSpec.scala
```diff
@@ -183,6 +183,17 @@ class OdinsonControllerSpec extends PlaySpec with GuiceOneAppPerTest with Inject
       println(Helpers.contentAsString(response))
     }
 
+    "respond with frequency table using the /term-hist endpoint" in {
+      val body = Json.obj("field" -> "lemma")
+
+      val response =
+        route(app, FakeRequest(GET, "/api/term-hist?field=lemma").withJsonBody(body)).get
```
myedibleenso, line 190: Can we add another test that includes some of the other optional params for this endpoint (ex. log scale)?

myedibleenso (changes requested):

danebell:
I have made the requested changes, and added two endpoints `rule-freq` and `rule-hist` to provide analogous statistical frequency tables to `term-freq` and `term-hist`, respectively.

myedibleenso:
Thank you, @danebell !

PR #263, opened 2021-03-02
Downgrade the pgp plugin for release

kwalcock:


kwalcock:
This version has worked in the past and is compatible with other projects.

kwalcock:
See also https://github.com/clulab/reach/pull/702

I didn't try too hard to find a solution, but this has been a problem before.

danebell (approved):
Looks good to me.

PR #267, opened 2021-03-16
Initiate ExtractorEngine with every API call that needs it

danebell:
This change prevents the mentions from one run's grammar showing up in a later call. This is important, because otherwise a user making multiple grammar calls will see irrelevant mentions in their responses. These changes are expensive because they prevent the caching which would otherwise speed up the EventExtractor.

A test is included to check event type and rule name exclusion.


BeckySharp:
if this is temporary, until something better is done, should the fix be to clear state instead?
Also -- maybe it would be possible to have a local State in the scope of the `extractMentions` method that either points at the session State (i.e., the one that lasts for the life of the EE) or points at an ephemeral one.  that way you could support BOTH the interactive use case and the issue with thread-safely.  ?

danebell:
> if this is temporary, until something better is done, should the fix be to clear state instead?
I can't think of an argument against that solution at the moment, but I thought @myedibleenso had some objection. Maybe two people make a call simultaneously, and there's leakage between the two?

> Also -- maybe it would be possible to have a local State in the scope of the `extractMentions` method that either points at the session State (i.e., the one that lasts for the life of the EE) or points at an ephemeral one. that way you could support BOTH the interactive use case and the issue with thread-safely. ?
That's a good idea. I don't see a problem with it.

myedibleenso:
>should the fix be to clear state instead?

Given the state is an attribute of the ee, simply clearing state is probably unsafe when requests are made in parallel.  Even if synchronized, it would mean waiting for one query to end before another could actually be executed.

>maybe it would be possible to have a local State in the scope of the extractMentions method that either points at the session State (i.e., the one that lasts for the life of the EE) or points at an ephemeral one. that way you could support BOTH the interactive use case and the issue with thread-safely. ?

> the one that lasts for the life of the EE

to support the interactive use case, you'll potentially need many such states (at least one per user).  Supporting that sounds like a larger change and one I'm not sure will be easy to manage.

>maybe it would be possible to have a local State in the scope of the extractMentions

Adopting this strategy would mean we could at least guarantee a clean state per call and we'd get the benefit of "warm" starts by avoiding needing to create a new ee instance for each request.  This change is a bit more involved, but if it's something we can do fairly quickly (and test) I think this is a much better solution and it might provide a mechanism for injecting user or session-specific states later on (managed elsewhere).

myedibleenso:
To clarify, making `state` a param of `extractMentions` and having it default to the existing one sounds like a better solution.

danebell:
> To clarify, making `state` a param of `extractMentions` and having it default to the existing one sounds like a better solution.

Sorry, I need further clarification. What is the type of the `state` param you're proposing? Is it a Boolean that says whether to clear the state before/after the call?

myedibleenso:
The type would be `ai.lum.odinson.state.State`, but after looking at it further I think this change would require _many_ other changes to ExtractorEngine in order to pipe the state through `extractFromPriority`, [`processAndPromote`](https://github.com/lum-ai/odinson/blob/16499562cb6bebf02531e1c6d16d6a4c7000f039/core/src/main/scala/ai/lum/odinson/ExtractorEngine.scala#L395-L400), [`handleArgumentPromotion()`](https://github.com/lum-ai/odinson/blob/16499562cb6bebf02531e1c6d16d6a4c7000f039/core/src/main/scala/ai/lum/odinson/ExtractorEngine.scala#L411-470), etc. to to [`query()`](https://github.com/lum-ai/odinson/blob/16499562cb6bebf02531e1c6d16d6a4c7000f039/core/src/main/scala/ai/lum/odinson/ExtractorEngine.scala#L192-L210)


One alternative would be to make `state` a `var`, but that would mean queries would have to wait their turn in order to execute (i.e., no parallelism in requests).  In other words, I don't think it's a quick fix to pass a state param to `extractMentions` (I still think this might be the way to ultimately go, though).

BeckySharp:
gus, the ability to pass in a state seems reasonable to me, and likely
something we could do quickly

BeckySharp:
and this could be done optionally -- that is, if you pass it, use that state, if you don't, use the EE state

danebell:
@myedibleenso @BeckySharp I don't know how to proceed.

myedibleenso:
@danebell , let's discuss tomorrow.  In the meantime (hopefully), I will deploy this branch for the sake of the demo service.

myedibleenso:
Merge conflcts!  Hip hip hooray!

BeckySharp (approved):
with the power vested in me by the opener of this PR, I approve this PR.

BeckySharp:
@myedibleenso @danebell @marcovzla 
I think we need to close the `IndexSearcher` when the API exits?
I had a residual write.lock after running things that I wasn't expecting.  The tests all pass so I didn't notice until I tried to run something else on a diff branch that is not from this work.

BeckySharp:
if you saw the thing i deleted -- false alarm!

BeckySharp:
@myedibleenso @danebell is this ready to merge?

myedibleenso:
I think we should remove those `println`s or switch them to log statements.

BeckySharp:
> I think we should remove those `println`s or switch them to log statements.

good catch, I think that @danebell said they were unintentionally left in, so @danebell do you want to remove them? or should i?

danebell:
I'll take care of it.

myedibleenso (commented):
I think this looks great, but apparently coverage is down just enough to fail that check...

BeckySharp:
in the past I don't think we've held ppl to that check, idk if that means
we should adjust it?

myedibleenso:
There are two coverage-related checks.  One has not be required (the patch one).  The other, which checks to see if the project's coverage has decreased, has been required.  We can change that if we want.

BeckySharp:
ok, i tried to get more detail than this, but seems I am not authorized.  Regardless, seems like there was one many "hit" to the coverage: 
image

Any idea what the `OdinsonFilteredQuery` coverage went down so much with this PR?

BeckySharp:
regardless, I'm inclined for us to not care too much about small variations in the coverage like this. (-0.79%)

marcovzla (approved):
LGTM

marcovzla:
I vote for merging as is.

PR #270, opened 2021-03-29
Reuse the IndexSearcher to keep caching advantages

BeckySharp:
This allows the Controller to maintain an IndexSearcher (the thing that is making use of the cacheing) and create each ExtractorEngine with it, so the speed-up is maintained in subsequent calls, but each call still gets its own extractorengine

FYI @marcovzla


myedibleenso:
Thanks, @BeckySharp , for taking the initiative with this!  I had the same idea in mind after a discussion last week.

myedibleenso (approved):

BeckySharp:
@myedibleenso I can't take any credit for the idea, @marcovzla shared the idea you guys discussed last week, just had some time

PR #274, opened 2021-03-30
Data gatherer

BeckySharp:
Let's discuss

### DIFF
--- a/core/src/main/scala/ai/lum/odinson/DataGatherer.scala
+++ b/core/src/main/scala/ai/lum/odinson/DataGatherer.scala
@@ -20,22 +20,10 @@ class DataGatherer(
 
   val analyzer = new WhitespaceAnalyzer()
 
-  val storedFields = indexSettings.storedFields
+  val storedFields: Seq[String] = indexSettings.storedFields
 
   def getStringForSpan(docID: Int, m: OdinsonMatch): String = {
     getTokensForSpan(docID, m).mkString(" ")
-  }
-
-  def getArgument(mention: Mention, name: String): String = {
-    getStringForSpan(mention.luceneDocId, mention.arguments(name).head.odinsonMatch)
-  }
-
-  def getTokensForSpan(m: Mention): Array[String] = {
-    getTokensForSpan(m.luceneDocId, m.odinsonMatch, displayField)
-  }
-
-  def getTokensForSpan(m: Mention, fieldName: String): Array[String] = {
-    getTokensForSpan(m.luceneDocId, m.odinsonMatch, fieldName)
   }
 
   def getTokensForSpan(docID: Int, m: OdinsonMatch): Array[String] = {
@@ -62,8 +50,6 @@ class DataGatherer(
     getTokens(scoreDoc.doc, fieldName)
   }
 
-  // TODO: deprecate this, and then remove -- this change is to preserve functionality
-  //   in the meantime
   def getTokens(docID: Int, fieldName: String): Array[String] = {
     TokenStreamUtils
       .getTokensFromMultipleFields(docID, Set(fieldName), indexReader, analyzer)(fieldName)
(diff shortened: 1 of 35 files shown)


myedibleenso:
>Let's discuss

@BeckySharp , can you please summarize this?

BeckySharp:
yes  @myedibleenso for sure!  

The overarching goal is to (across a few PRs) make the Mentions much easier to work with _after_ extracting.  
OK, so the idea is that a "datagatherer" is the guy who is able to get the annotation content of the mentions (e.g., raw, lemmas, tags, entities, etc.).. In this sense, the ExtractorEngine is no longer responsible for this directly (clarifying "responsibility").

Going right along with this -- the idea is that initially, extracted mentions are *minimal* (like they currently are), they don't know their text, their containing sentence text, their linguistic annotations etc.  Then, if you want to (as many downstream people do) you can `populate` them -- basically *add this content* so that the Mentions themselves can tell you things you want to know about them (text etc.).  To try and maintain the efficiency of things, there are different levels of population -- (a) none (i.e., don't populate), (b) `display` (displayField only, for both the mention and the containing document/sentence), and (c) `all`, (all stored fields available in the index for both the mention and the containing document/sentence).

This PR has the decoupling, has the attributes of the Mention, and has basic access methods in the Mention (`mention.text`), but in a _later_ PR I plan to add more access methods to add some of the much-loved ease of use of Odin Mentions.

The decision to use `private var` in the Mention was to avoid creating extra objects (and so to make it faster/lighter weight), however, that behavior can be modified.  If not modified, fwiw, from the outside they behave as immutable.

Please ask more questions, give comments, etc

core/src/main/scala/ai/lum/odinson/DataGatherer.scala
```diff
@@ -0,0 +1,130 @@
...
+  def getTokens(scoreDoc: OdinsonScoreDoc): Array[String] = {
+    getTokens(scoreDoc.doc, displayField)
+  }
+
+  def getTokens(scoreDoc: OdinsonScoreDoc, fieldName: String): Array[String] = {
+    getTokens(scoreDoc.doc, fieldName)
+  }
+
+  def getTokens(docID: Int, fieldName: String): Array[String] = {
+    TokenStreamUtils.getTokens(docID, fieldName, indexSearcher, analyzer)
```
BeckySharp, line 92: per marco -- try to intelligently load only fields we need:
https://lucene.apache.org/core/6_6_6/core/index.html

BeckySharp:
`val c = b.map{x => x.append(2);x}`

BeckySharp:
With this PR now -- Mentions can populate themselves with their lexical info (they have a pointer to a DataGatherer).
 --> `Mention.populateFields`
This includes the lexical info for the mention itself as well as the Document (i.e., sentence)

Also, the `ExtractorEngine` has a method for getting mentions called `extractAndPopulate` which returns an Iterator of mentions that are already populated.  This makes it one less step for the consumer if they want that.

Per request, TokenStreamUtils was modified to only fetch the fields desired from the index to save time.

Tests are added in TestMention and TestTokenStreamUtils

in an upcoming PR I plan to deprecate some `getTokens` methods in the ExtractorEngine and DataGatherer.

BeckySharp:
fyi @myedibleenso and @danebell -- this may make things easier in the backend?  If adjustments are needed for your purposes also, please let me know! 
thanks!

core/src/main/scala/ai/lum/odinson/lucene/analysis/TokenStreamUtils.scala
```diff
@@ -2,23 +2,41 @@ package ai.lum.odinson.lucene.analysis
...
+  def getTokensFromMultipleFields(
     docID: Int,
-    fieldName: String,
+    fieldNames: Set[String],
     indexSearcher: IndexSearcher,
     analyzer: Analyzer
-  ): Array[String] = {
-    val doc = indexSearcher.doc(docID)
+  ): Map[String, Array[String]] = {
+    val doc = getDoc(docID, fieldNames, indexSearcher)
```
BeckySharp, line 27: just need a reader? if so, only use reader above in the datagatherer

marcovzla (approved):
Thanks!

PR #281, opened 2021-04-18
draft of parent query documentation

maxaalexeeva:
- added information about parent queries to documentation

Note: the link to the test suite is what I expect it to be in master when it's merged.

### DIFF
--- a/docs/parent_queries.md
+++ b/docs/parent_queries.md
@@ -1,0 +1,40 @@
+---  
+title: Parent Queries
+parent: Queries
+has_children: false
+nav_order: 5
+---  
+
+# Parent queries
+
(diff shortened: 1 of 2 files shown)


kwalcock:
That CLA might not apply to you.  We haven't gotten around to making an exception list.  I'll assume that someone will tell me if it is necessary.

maxaalexeeva:
@kwalcock, ah, thanks! I was mid-reading it, but I will pause on that until further notice then.

kwalcock:
@maxaalexeeva, it looks like the exception list works.  You do not need that agreement.  Thanks.

maxaalexeeva:
@kwalcock, thanks for checking!

PR #283, opened 2021-04-21
Binary

BeckySharp:
Per request, switch to BinaryDocumentFields for the dependency graphs, allows for larger graphs.

@marcovzla a test is missing


BeckySharp:
@marcovzla I think this PR may have made it so indexes generated here are incompatible with older ones.  Should we then publish with a bumped major version?

PR #285, opened 2021-05-10
Improve efficency of the term-freq endpoint

danebell:
The previous implementation of term-freq (the API endpoint for returning graphable JSON tables of term frequency in a dataset) was inefficient to the point of timing out rather than returning for even basic queries. The following corrections improve this performance:

1. Use Lucene's native `RegExp` and `Automaton` classes to filter terms rather than checking each term with Scala regex.
2. Create a `FrequencyTable` class for iteratively tracking the highest-ranked (by lowest or highest frequency order, as requested) terms. Use a `Queue`-like `List` structure, inserting new terms that are higher-ranked than the current boundary and pushing out terms beyond the `max` requested index. This means that sorting is done as terms are examined, and there is no additional need to sort a large `List` of terms after retrieving frequencies.
3. When the API call requests alphanumeric ordering, take terms off of the `TermEnum` in sequence until we have `max +1` of them. This is possible because `TermEnum` uses alphanumeric (Unicode sort) order. Unfortunately, this means that if the API call also requests reverse order (i.e. last alphanumeric first), the system needs to iterate through all the terms, but as it only needs to save the last `max + 1` items, this action is still speedy.

The relevant issue is https://github.com/lum-ai/odinson-ide/issues/53

### DIFF
--- a/backend/app/controllers/OdinsonController.scala
+++ b/backend/app/controllers/OdinsonController.scala
@@ -118,7 +118,7 @@ class OdinsonController @Inject() (config: Config = ConfigFactory.load(), cc: Co
 
   }
 
-  /** For a given term fied, find the terms ranked min to max (inclusive, 0-indexed)
+  /** For a given term field, find the terms ranked min to max (inclusive, 0-indexed)
     * @param field The field to count (e.g., raw, token, lemma, tag, etc.)
     * @param group Optional second field to condition the field counts on.
     * @param filter Optional regular expression filter for terms within `field`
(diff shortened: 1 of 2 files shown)


danebell:
After consideration, I'm going to make all instances of regex unanchored. I think this more closely matches what the user will expect, e.g. when filtering by `the`, the user expects to get `the`, `there`, `other`, etc.

danebell:
Having tilted at windmills for a while, Lucene's RegExp is so specialized that I ended up with the following (hopefully temporary) compromise:
- all filters are anchored (i.e. must match the whole string)
- `termFreq` uses Lucene's RegExp (subset of Perl-style regex) because it's fast that way
- `ruleFreq` uses Scala-style regex because it's too hard to use RegExp for anything but Lucene stuff

With apologies to any users who are subsequently confused.

PR #287, opened 2021-05-12
first pass at incremental indexing

reynoldsm88:
@marcovzla @BeckySharp @danebell 

I took a pass at adding incremental indexing to the `OdinsonIndexWriter`. Based on the unit tests this seems to work in theory, however, I have not tested it out in a real application yet. I'll update this thread when I have more information in that regard. The changes are fairly straightforward, however, I'm going to mark one particular change that I need someone with more expertise to evaluate if I did something untoward or not.

### DIFF
--- a/core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala
+++ b/core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala
@@ -50,8 +50,8 @@ class OdinsonIndexWriter(
 
   val writerConfig = {
     val writerConf = new IndexWriterConfig(analyzer)
-    if (!incremental) writerConf.setOpenMode(OpenMode.CREATE)
-    else writerConf.setOpenMode(OpenMode.CREATE_OR_APPEND)
+    if (incremental) writerConf.setOpenMode(OpenMode.CREATE_OR_APPEND)
+    else writerConf.setOpenMode(OpenMode.CREATE)
     writerConf
   }
 
@@ -74,7 +74,7 @@ class OdinsonIndexWriter(
   }
 
   def indexDocuments(block: Collection[lucenedoc.Document]): Unit = {
-    writer.addDocuments(block)
+    addDocuments(block)
     if (incremental) commit()
   }
 
--- a/core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala
+++ b/core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala
@@ -25,7 +25,7 @@ import ai.lum.odinson.serialization.UnsafeSerializer
 import ai.lum.odinson.utils.IndexSettings
 import ai.lum.odinson.utils.exceptions.OdinsonException
 import org.apache.lucene.document.BinaryDocValuesField
-import ai.lum.odinson.{Document => OdinsonDocument}
+import ai.lum.odinson.{ Document => OdinsonDocument }
 
 class OdinsonIndexWriter(
   val directory: Directory,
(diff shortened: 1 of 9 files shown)


core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala
```diff
@@ -81,9 +101,16 @@ class OdinsonIndexWriter(
     addDocuments(block)
   }
 
-  def commit(): Unit = writer.commit()
+  def commit(): Unit = {
+    writer.flush()
+    writer.commit()
+  }
 
   def close(): Unit = {
```
reynoldsm88, line 109: @marcovzla - can you review the changes to this method? The Lucene methods don't appear to support overwriting or merging existing files, so the only way I could get it to work was to rewrite the file each time...

core/src/test/scala/ai/lum/odinson/foundations/TestIndexWriter.scala
```diff
@@ -72,7 +71,58 @@ class TestOdinsonIndexWriter extends OdinsonTest {
...
+    // add one doc...
+    val docOne : OdinsonDocument = getDocument("alien-species")
+    indexer.addDocument(docOne)
+
+    reader = DirectoryReader.open(FSDirectory.open(indexDir.toPath))
+
+    reader.numDocs shouldBe 2 // number of lucene entries, not documents
+    reader.close()
+    indexer.close()
+
```
BeckySharp, line 113: @reynoldsm88 to really test this -- can you add a line here that (a) runs a query, you get N results, (b) then you add the new document, (c) and with the same extractor engine, run the same query and you get N + M results?

core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala
```diff
@@ -81,9 +101,16 @@ class OdinsonIndexWriter(
...
+    if (directory.listAll().contains(VOCABULARY_FILENAME)) directory.deleteFile(VOCABULARY_FILENAME)
```
BeckySharp, line 110: can you please add a small comment here to help ppl know why these files are being deleted? thanks!

BeckySharp (changes requested):
Just went over this with @marcovzla -- looks great!!
We left two small comments/requests amidst our gratitude :)
Also -- travis will cry unless you run the linting step -- 
`sbt scalafmtAll && sbt scalafmtCheckAll`
(http://gh.lum.ai/odinson/contributing.html)

BeckySharp:
this is great!

reynoldsm88:
https://github.com/lum-ai/odinson/pull/287#discussion_r631374601

@BeckySharp @marcovzla - I'm using two randomly selected documents from `ai.lum.odinson.utils.TestUtils.ExampleDocs`. It looks like those are already parsed into Odinson json so it's hard to tell what the actual text is. Is there a generic rule you normally use that will work for any document?

reynoldsm88:
@BeckySharp @marcovzla so i did some investigating into how `Lucene` internally handles incremental reads and writes. It turns out that there is a [special class provided by Lucene](https://lucene.apache.org/core/6_6_6/core/org/apache/lucene/search/SearcherManager.html) that is used to wrap and manage the lower level API's like `IndexWriter` and `IndexSearcher` for this very purpose.

I've stubbed out a POC using this `Lucene` API that shows how to do incremental indexing. There is a test that validates the approach. Note this is only from a `Lucene` perspective, we would need to integrate these changes to `Odinson` to validate that it will work with an `ExtractorEngine`:
[IncrementalLuceneIndex](https://github.com/reynoldsm88/odinson/blob/797456eb9823be62091d6d5b1f4ea13cba4f44f1/core/src/main/scala/ai/lum/odinson/lucene/LuceneIndex.scala#L30)
[TestLuceneIndex](https://github.com/reynoldsm88/odinson/blob/797456eb9823be62091d6d5b1f4ea13cba4f44f1/core/src/test/scala/ai/lum/odinson/lucene/TestLuceneIndex.scala#L15)

Based on my findings, my suggestion would be to refactor some of the code such that the `Odinson` classes do not interact with low level `Lucene` API's such as `IndexSearcher`. Instead, I would propose that we wrap `Lucene` functionality into an interface similar to what I have stubbed out here. In this way, we can create multiple implementation and include one that properly delegates functionality to the incremental indexing utilities.

I am happy to work on this, but I would like you to review the proof of concept I have stubbed out to make sure you understand and agree with my proposal.

PR #292, opened 2021-05-25
Tag vocabulary

danebell:
These changes create a new endpoint, `/api/tags-vocabulary` analogous to the existing endpoint `/api/dependencies-vocabulary`. Because there is no file like `dependencies.txt` in the index, this is achieved by looking at the `Fields` of the `IndexReader` and picking out the `Term`s of the `Field` with the same name as the config for `posTagTokenField`. These are returned (in Lucene's orthographic UTF-8 order) to the user as a JSON array of strings.

Tests for this new endpoint are included, as well as tests for the `/api/dependencies-vocabulary` endpoint and the `/api/corpus` endpoint.


myedibleenso:
@danebell , I'm wondering if this information should be included in the JSON file introduced in #255 .  Thoughts?

BeckySharp:
> @danebell , I'm wondering if this information should be included in the JSON file introduced in #255 . Thoughts?

depending on size of the list, makes a lot of sense to me, fwiw, else it could also go into a separate file like tags.txt or something.  If these types of files seem like they will proliferate (which we should try to semi-prevent prob), we can put then in an `index_info` subdir (not deps prob, bc it's directly used by the index, but others).

danebell:
I'd be even happiest if we could have a _complete_ list of possible tags from the annotator of choice. That way it'd be complete and invariant even when new documents are added. Whether we keep it in the JSON or in a separate tags.txt file doesn't make much of a difference to me -- I think there are few enough tags to fit comfortably in the JSON. More relevant than size is that there might be additional discrete collections users might want, and we should have a policy that works for all of them.

Can we try to enumerate all the things we might want to store this way? Beyond dependencies and tags, I can think of NER labels (e.g. `B-ORG`) , chunk labels (e.g. `B-NP`), and semantic role labels (e.g. `A0`). Anything else come to mind? Is this an open-ended list? I can conceive of someone wanting spell check or similar for raw/words/lemma, but obviously that would be much more cumbersome.

kwalcock:
@danebell, this may be completely off the mark, but you may want to check the file TagSet.scala in recent versions of processors for some of them.

BeckySharp:
I like where this is going. I think that a full list of potential tags is great, and I like the idea that it would be invariant, but I don't know that that is possible (see below).

I'd also like info on _what annotator_ (type and version) was used, so we still know after the fact.
we could add something to the config in the `index` section that specifies which fields get their vocabs listed out, and the indexwriter builds a vocab for them?
That said, this solution would _not_ be based on the annotator, instead it would be based on the corpus.... i think that with all the different options for parsers out there we likely won't be able to have the corpus-invariant version...

What else... I still plan to add the size of the index (in sentences if not docs)... thoughts?

danebell:
> @danebell, this may be completely off the mark, but you may want to check the file TagSet.scala in recent versions of processors for some of them.

That's helpful, thanks. I'd prefer it if we could get it from the annotator itself upon creation, so that if Processors (or some equivalent) takes on new languages or syntactic models, the process of collecting them would be the same.

> I'd also like info on _what annotator_ (type and version) was used, so we still know after the fact.
Good idea. I think that'd be helpful. And it could be added to the `/api/buildinfo` endpoint, which currently returns
```
{
  "builtAt": "2021-03-23T14:58:14+0000",
  "name": "odinson-core",
  "gitUncommittedChanges": "true",
  "gitHeadCommit": "27dc4a50f08b4d853aaf44bc1a76d039ac513788",
  "gitCurrentBranch": "singleEventExtractor",
  "scalaVersion": "2.12.10",
  "version": "0.3.2-SNAPSHOT",
  "sbtVersion": "1.2.6",
  "gitHeadCommitDate": "2021-03-22T15:17:00-0400"
}
```

> we could add something to the config in the `index` section that specifies which fields get their vocabs listed out, and the indexwriter builds a vocab for them?
Very good idea! I like that solution a lot. Much like the setting for which fields to save.

> That said, this solution would _not_ be based on the annotator, instead it would be based on the corpus.... i think that with all the different options for parsers out there we likely won't be able to have the corpus-invariant version...
:-( That makes sense, but it's disappointing. Anyway, a corpus-level solution is still pretty good.

> What else... I still plan to add the size of the index (in sentences if not docs)... thoughts?
I guess the index size will have to be updated as new docs are added, right? That information seems useful for a human to have readily available. No other ideas right now about what else to add.

myedibleenso:
>we likely won't be able to have the corpus-invariant version...


I don't think we should even try for that.

>I guess the index size will have to be updated as new docs are added, right?

For what it's worth, we already have an endpoint that computes the sentence count on-the-fly.

myedibleenso:
@danebell , can you address the conflicts in `OdinsonController`?

danebell:
> > we likely won't be able to have the corpus-invariant version...
> 
> I don't think we should even try for that.

Can you say more about that? My motivation is that if you're correcting the POS tags on a small corpus, you might conceivably not have all the tags in a given scheme (or semantic roles or entities or whatever) extant in the index. I don't mean that the list should be invariant from corpus to corpus, but rather that adding additional documents should ideally not change the range of values allowed for tags (or whatever field).

danebell:
So if I can briefly summarize my understanding, the `OdinsonIndexWriter` currently writes a copy of the `Vocabulary` upon `close`. What's being proposed is that what gets written in addition (e.g. tags, chunks, etc.) to this should be specified in the `core` config. (Incremental indexing will soon mean that we'll need to overwrite these collections by manually deleting the existing files and writing them as if new upon every `close` call.)  Once this is accomplished, the `tags-vocabulary` endpoint can stay in place but be implemented as merely recalling the contents of these file. Maybe we should consolidate these into a single `vocabulary` endpoint to which one passes a string specifying the vocabulary (including `dependency`) desired.

BeckySharp:
per meeting: caching can happen in diff places, but look at how to cache an entrypoint/function through scala play

backend/app/controllers/OdinsonController.scala
```diff
@@ -733,6 +737,46 @@ class OdinsonController @Inject() (config: Config = ConfigFactory.load(), cc: Co
...
+  /** Return all terms for a given field in orthographic order.   *
+    * @param field A token field such as word, lemma, or tag.
+    * @return The complete [[List]] of terms in this field for the current index.
+    */
+  private def fieldVocabulary(field: String): List[String] = {
+    // get terms from the requested field (error if it doesn't exist)
+    val extractorEngine: ExtractorEngine = newEngine()
+    val fields = MultiFields.getFields(extractorEngine.indexReader)
+    val terms = fields.terms(field)
+    val termsEnum = terms.iterator
```
BeckySharp, line 749: per @kwalcock : there exists a converter from a java iterator to a scala iterator (look in collections.javaconverters._ or something)
then just to list

danebell:
Okay, @myedibleenso & @BeckySharp, now caching is ready for the `tags-vocabulary` endpoint, and the timeout is configurable. I set it to 1 hour by default.

backend/app/utils/LuceneHelpers.scala
```diff
@@ -0,0 +1,35 @@
+package utils
```
danebell, line 1: TODO: Can this be moved to core?

kwalcock:
Those are the most beautiful docstrings I have ever seen!  Thanks for adding them.

myedibleenso:
Looks like some backend tests are now failing after the change of formats to some of the fields used in the controller.

myedibleenso (changes requested):
Some tests are failing in the `backend` subproject...

myedibleenso (approved):

PR #297, opened 2021-06-02
changes and version

BeckySharp:
We forgot to do this back on #283  👎 

The increase in the major version is bc indexes made prior are no longer compatible.


marcovzla (approved):
Nice!

PR #299, opened 2021-06-09
Upgrade processors and allow getting a document from a File's String contents

BeckySharp:
FYI @myedibleenso @kwalcock @danebell


BeckySharp:
fwiw @marcovzla the tests pass, it's the coverage diff that's sending the red x

PR #304, opened 2021-06-22
Make parent document's `fileName` a `storedField` by default

### LINKED ISSUE #303: add `fileName` to `savedFields` array in `core`'s config file

danebell:
In the Odinson API, multiple endpoints rely on the parent file being available to return or consume. This is only possible if the `parentDocFieldFileName` is stored in the index. This fact was undiscovered because there were no tests for these endpoints. These changes add `parentDocFieldFileName` to the `storedFields`, add simple tests for the 4 endpoints in question, and edit the `storedFields` comment to encourage the use of `parentDocFieldFileName`.

Closes #303

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
+- Added tests for metadata and parent API calls in backend.
 ### Changed
+- Parent document filenames are stored by default.
 
 ## [0.4.0] - 2021-06-07
 ### Added
@@ -14,12 +16,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Enhanced term-freq endpoint to allow filtering as well as grouping by a second field.
 - Added ability to Mentions to populate their lexical content ([#274](https://github.com/lum-ai/odinson/pull/274))
 - Added tests for parent queries in core and backend.
-- Added tests for metadata and parent API calls in backend. 
 ### Changed
 - Dependencies now stored as BinaryDocValuesField (previously SortedDocValuesField) to allow for larger graphs ([#283](https://github.com/lum-ai/odinson/pull/283)).
 - Moved responsibility for getting lexical content from ExtractorEngine to DataGatherer ([#274](https://github.com/lum-ai/odinson/pull/274))
 - Metadata is now indexed as TokensFields instead of StringFields.
-- Parent document filenames are stored by default.
 
 ## [0.3.0] - 2021-02-18
 ### Added 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,8 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
-- Added tests for metadata and parent API calls in backend.
 - Added tags-vocabulary endpoint to API for index-specific part-of-speech tags.
+- Added tests for metadata and parent API calls in backend.
 ### Changed
 - Parent document filenames are stored by default.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,10 +14,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Enhanced term-freq endpoint to allow filtering as well as grouping by a second field.
 - Added ability to Mentions to populate their lexical content ([#274](https://github.com/lum-ai/odinson/pull/274))
 - Added tests for parent queries in core and backend.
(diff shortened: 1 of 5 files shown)


BeckySharp:
@danebell should we enforce it too? like here:
https://github.com/lum-ai/odinson/blob/master/core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala#L283

or is that a bad idea?
if we don't enforce it, we should maybe change the "must" for that one to something like "should contain, except in extenuating circumstances" or something....

thoughts?

BeckySharp:
> @danebell should we enforce it too? like here:
> https://github.com/lum-ai/odinson/blob/master/core/src/main/scala/ai/lum/odinson/OdinsonIndexWriter.scala#L283
> 
> or is that a bad idea?
> if we don't enforce it, we should maybe change the "must" for that one to something like "should contain, except in extenuating circumstances" or something....
> 
> thoughts?

@danebell it could also be enforced in the backend -- for example, the `storedFields` value is written in the index settings file, so the backend could consult that and complain if it wasn't stored?

danebell:
I guess I'm torn. It's conceivable that someone could create an index and never use the API at all, right? But if that field goes missing for someone who does need to run the API, they need to know what's going wrong. The backend isn't invoked at all (AFAIK) in indexing, so it would have to complain only after you'd done your indexing and are trying to start up the API, right? 

What I want is for a person who intends to use the API to have prior warning that some endpoints won't work without the filename, but not impose that requirement on anyone planning to use Odinson in other ways.

BeckySharp:
> @BeckySharp I guess I'm torn. It's conceivable that someone could create an index and never use the API at all, right? But if that field goes missing for someone who does need to run the API, they need to know what's going wrong. The backend isn't invoked at all (AFAIK) in indexing, so it would have to complain only after you'd done your indexing and are trying to start up the API, right?

true
 
> What I want is for a person who intends to use the API to have prior warning that some endpoints won't work without the filename, but not impose that requirement on anyone planning to use Odinson in other ways.

yes -- not all users will use the API, some will use it programmatically as a library.

> Maybe the indexing could issue an informative warning if the filename isn't stored.

that makes sense to me.  maybe before we merge this we/you add (if these aren't already done):
- check in the controller or wherever that checks the settings.json file to make sure the filename was one of the stored fields and if not issues an appropriate error message
- add the warning in the IndexWriter?

myedibleenso:
>it could also be enforced in the backend -- for example, the storedFields value is written in the index settings file, so the backend could consult that and complain if it wasn't stored?

This is a good compromise.  Aside from logging a warning whenever an endpoint is called when this is missing, we could also add a field to the response for the /corpus endpoint indicating whether or not this field is stored:

https://github.com/lum-ai/odinson/blob/cdaa89cb0a763443db8198eee5b401fe685ab6ef/backend/conf/routes#L39
https://github.com/lum-ai/odinson/blob/cdaa89cb0a763443db8198eee5b401fe685ab6ef/backend/app/controllers/OdinsonController.scala#L696-L700

myedibleenso (commented):
This looks good, but I think we should discuss https://github.com/lum-ai/odinson/pull/304#issuecomment-866220278

danebell:
@myedibleenso 
> This is a good compromise. Aside from logging a warning whenever an endpoint is called when this is missing, we could also add a field to the response for the /corpus endpoint indicating whether or not this field is stored:

How about a key showing all the `storedFields`? It isn't as explicit about the need for `parentDocFieldFileName` to be stored, but I don't think a key for that would be very clear to most users anyway.

@BeckySharp 
> * check in the controller or wherever that checks the settings.json file to make sure the filename was one of the stored fields and if not issues an appropriate error message

I agree this is a good idea. Unsure of what the appropriate error message would be. I think the current error message is very uninformative. Should this be a server-side error that halts the API starting, or a client-side 500 status code showing that the server is misconfigured?

myedibleenso:
>How about a key showing all the storedFields? It isn't as explicit about the need for parentDocFieldFileName to be stored, but I don't think a key for that would be very clear to most users anyway

Yeah, that sounds good.  Do you think `/corpus` is that the right endpoint?  Or should we have  something like `/corpus/details`?

>Should this be a server-side error that halts the API starting, or a client-side 500 status code showing that the server is misconfigured?

I think it should be a server-side error, but it shouldn't prevent the API from starting.  Not every endpoint relies on that field.  For those that do, you'd want an informative error returned.

danebell:
> Yeah, that sounds good. Do you think `/corpus` is that the right endpoint? Or should we have something like `/corpus/details`?

I think it'd be fine in `/corpus`, but I'm not too clear on the use cases, so I could be wrong. It's not that much more information to pass along.

> I think it should be a server-side error, but it shouldn't prevent the API from starting. Not every endpoint relies on that field. For those that do, you'd want an informative error returned.

Okay, that makes sense.

PR #306, opened 2021-06-22
Use utils classes an implicits to iterate through Lucene

kwalcock:


kwalcock:
This might be useful here and elsewhere.  It has not been tested.  Perhaps it can be verified by someone who knows how to run the backend.

BeckySharp:
OK @kwalcock , so l love this, but also I don't get it :)
Where are any of these methods called in the `OdinsonController`?
Or is it like a super sneaky ninja implicit that is called behind the scenes somewhere in the.... `map`...? ???

kwalcock:
terms has iterator(), so it's an `Iteratorable`.  iterator() on terms returns a TermsEnum which has a next() and is therefore a Nextable.  So, terms is an Iteratorable[Nextable[T]] which can be converted implicitly into a Scala Iterable[T] and used in "normal" Scala operations.  This all happens before .map(_.utf8ToString).toList is called.  The compiler seems to think this, at least.  I haven't seen it happen at runtime.  I looked through the code for instances of != null and .next() and it seemed like it might be useful in a couple of places.

BeckySharp:
thanks for the explanation!
you're awesome!

BeckySharp:
@danebell do you have tests that can check if this works for you?

danebell:
> @danebell do you have tests that can check if this works for you?

@BeckySharp  Yes, [this test](https://github.com/lum-ai/odinson/blob/89ab438e5659e249c1e10f6b19ab6325edddcc46/backend/test/controllers/OdinsonControllerSpec.scala#L498-L507) should detect if it didn't work. My only concern is that @kwalcock's excellent implicits could be used elsewhere in that Controller. So maybe I should see where else it could be used first before merging.

kwalcock:
Don't let it hold you back.  If used at all, it can be merged instead into master in a separate operation after your merge.  It was just dependent on your changes so had to attach to them somehow.  Another thing to search for is the ArrayBuffer which might be involved.

PR #311, opened 2021-07-01
Use a custom Traversable for terms and freqs

kwalcock:
Address the cursor positioning of the iterator in LuceneHelpers, which is now unused.


BeckySharp (approved):
thanks!!

PR #249, opened 2021-02-05
Parent query bug fix

myedibleenso:
Parent queries (i.e., queries involving a filter on the fields of the parent document) quietly broke at some point in the past.  There was an inconsistency with the fields used by the `QueryCompiler` and `OdinsonIndexWriter` to identify parent document information which has been addressed here by introducing new config entries for those field names.

# Summary of changes
- new config entries for parent query fields
- improved consistency between the `QueryCompiler` and `OdinsonIndexWriter`
- parent query tests in the `backend` subproject

# TODOs
- [ ] Update documentation with a reference to [an example](https://github.com/clu-ling/odinson-example) demonstrating how to use parent queries via the REST API
- [ ] Update CHANGES
- [x] Add tests to `core`

PR #316, opened 2021-07-14
Parent file location field is stored by default

danebell:
The REST API requires the parent filename field to be stored, so store it by default.

### DIFF
--- a/core/src/main/resources/reference.conf
+++ b/core/src/main/resources/reference.conf
@@ -76,8 +76,11 @@ odinson {
 
     # list of document/sentence fields to store in index, **must** include the displayField
     storedFields = [
-        ${odinson.displayField}
+        ${odinson.displayField},
+        ${odinson.index.parentDocFieldFileName}
     ]
+
+    parentDocFieldFileName = fileName
 
     # the raw token
     rawTokenField = raw


BeckySharp (commented):
@danebell this was added to `extra/.../application.conf` in https://github.com/lum-ai/odinson/pull/249
should we close this?

dane:
Probably didn't mean me, but 👍 to closing

BeckySharp:
> Probably didn't mean me, but 👍 to closing

haha you are right, thanks and sorry for the wrong number!

danebell:
@BeckySharp This **should** have been added in #249 but wasn't due to an oversight. It needs to be in core, too.

BeckySharp:
> @BeckySharp This **should** have been added in #249 but wasn't due to an oversight. It needs to be in core, too.

It doesn't need to be in core, I think last we all talked we decided core should be agnostic to the format of the document (e.g., is it a file? a website, etc)
I can be in the backend though, and is now, as of #320 

Or am I missing something?
It's also in `extra`, for when building the index

danebell:
Oh, I didn't see that #318 is resolved. It seems that this can be closed, then.

PR #324, opened 2021-08-04
Metadata grammars

BeckySharp:
Added support for putting metadata filter queries in the grammars, either as a single string or as a list to be combined with AND

Needed still are something in the docs.
Also, please review and let me know if things need changing


BeckySharp:
also -- sorry, I meant to make the test in core/src/test/scala/ai/lum/odinson/events/TestEvents.scala a diff PR but accidentally committed it here, but it's just another test of event syntax for good measure, this one where the trigger is like `@PrevFound`

marcovzla (approved):
Looks great! Thanks!

Issue #328, opened 2021-08-18
Unable to filter using metadata


victoryhb:
I am experimenting with a custom annotation pipeline to generate Odinson documents for indexing. After some trial and error (since I couldn't find related info in the docs:), I was able to index the documents along with some metadata. The current metadata of one of my files as shown via Odinson's REST API interface (/api/metadata/by-document-id, version 0.5.1 Snapshot) is something like the following:
```
[
  {
    "$type": "ai.lum.odinson.StringField",
    "name": "file",
    "string": "AAT"
  },
  {
    "$type": "ai.lum.odinson.StringField",
    "name": "corpus",
    "string": "BNC"
  }
]
```

However, when trying to filter documents with the filter: corpus=='BNC',  either using the parentQuery of /api/execute/pattern or defining the metadataFilters parameter through /api/execute/grammar,  no results were returned (the results were correctly returned without the filter). Below is the grammar I tested with:

```
metadataFilters: 
 - corpus=='BNC'

rules:
 - name: "example"
   label: GrammaticalSubject
   type: event
   pattern: |
     trigger = [lemma=have]
     subject  = >nsubj []
```

I wonder if I am doing anything wrong here?

BeckySharp:
@victoryhb thanks, I'll work on reproducing this tonight.  BTW, there should be a PR very soon that will make it easier to add metadata (I thought I had merged it tbh, but i don't see it, which means I must have forgot to finish some last step...)

BeckySharp:
> @victoryhb thanks, I'll work on reproducing this tonight. BTW, there should be a PR very soon that will make it easier to add metadata (I thought I had merged it tbh, but i don't see it, which means I must have forgot to finish some last step...)

ha! i'm an idiot, it _was_ merged in #319 though you're right that there isn't documentation about it...

BeckySharp:
@victoryhb reproduced.  However, it's not a bug per se as much as poor documentation.
The StringField is not supported for metadata.
If you convert your metadata to this:
```
[
  {
    "$type": "ai.lum.odinson.TokensField",
    "name": "file",
    "tokens": ["AAT"]
  },
  {
    "$type": "ai.lum.odinson.TokensFIeld",
    "name": "corpus",
    "tokens": ["BNC"]
  }
]
```

it should work.  I will:

1. add a test for this advice
2. add an exception for StringFields as metadata
3. add info to the documentation

victoryhb:
[...]

PR #329, opened 2021-08-19
Documentation about StringField unsupported for metadata

BeckySharp:
Addresses #328 

Documentation failed to explain that StringFields are not supported in the Odinson Document metadata.  @marcovzla  can correct me if I'm wrong, but I think it has to do with how we build Lucene queries against text (i.e., to support the contains queries etc.)
If there is substantial need for supporting this, we can reconsider.
In the meantime, users should consider reframing their metadata as another field type (e.g., TokensField).
Note that this implies that users will tokenize their text. That is, "Biology Magazine" should not be represented as a list of one token that is `"Biology Magazine"`, but rather as `["Biology", "Magazine"]` (or another tokenization).

Added:
- info to the Docs pages
- a logger info statement that is logged if you (a) try to index a Document with user-defined StringField in the metadata or (b) try to use the `addMetadata()` method in `Document` class with a StringField.
- tests demonstrating the right and wrong behavior


BeckySharp:
also edited one of the example doc jsons (bad-character) to use TokensFields instead of StringFields

BeckySharp:
Also, please note comment here:
https://github.com/lum-ai/odinson/issues/328#issuecomment-901600227

do we need to update something(s) in the backend?

myedibleenso:
@BeckySharp , you asked about the REST API.  I think that minimally the following changes should be made:

Remove this line:
https://github.com/lum-ai/odinson/blob/3c5326be908e0648120f329e287c8667af0df4ce/backend/app/controllers/OdinsonController.scala#L857

Refactor this code to not use a parent query and possibly parse/compile the rule file in a way that is compatible with your changes to parent/metadata queries:
https://github.com/lum-ai/odinson/blob/3c5326be908e0648120f329e287c8667af0df4ce/backend/app/controllers/OdinsonController.scala#L881-L894

For consistency, @danebell, [`parentQuery` in the OpenAPI spec should be replaced with `metadataQuery`.  There's also a section describing `parentQuery` as a Lucene query.  This is no longer true.  Really we should **include a link to documentation describing how metadata queries work (what is supported, etc.)**. ](https://github.com/lum-ai/odinson/blob/master/backend/public/schema/odinson.yaml)

myedibleenso:
@BeckySharp @danebell , you can disregard my last message. I've taken care of those changes.

myedibleenso (approved):

PR #333, opened 2021-08-21
Patch/stored fields

BeckySharp:
By mixing the metadata stored fields (e.g., filename) and the sentence stored fields (e.g., raw, etc) there was an issue when the Mention tried to populate itself to the `VerboseLevels.all` level.  Basically, the mention would try to populate the metadata stored fields, but the lucene doc for the sentence didn't have that info (bc it was in a diff parent-level document since it's metadata).  This resulted in an exception.

This patch fixes that by separating the two and using them in the right places.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - `parentQuery` -> `metadataQuery` in REST API docs
 - Added documentation about supported types of fields for metadata (and info warnings for unsupported types). ([#329](https://github.com/lum-ai/odinson/pull/329))
 ### Changed
+- Fixed issue where the metadata stored fields caused Mention.populate to crash ([#333](https://github.com/lum-ai/odinson/pull/333))
 
 ## [0.5.0] - 2021-08-07
 ### Added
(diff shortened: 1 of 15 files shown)


BeckySharp:
let me fix the failing tests and reopen

BeckySharp:
fixed the tests, I had failed to update usage of `sentenceStoredFields` in the tests

myedibleenso:
It seems like this change will further complicate the config and quietly break old configs by breaking stored fields into two categories.  Can we not simply ignore stored fields that are missing from the sentence when making mentions?

BeckySharp:
> It seems like this change will further complicate the config and quietly break old configs by breaking stored fields into two categories. 

I def agree with this point

> Can we not simply ignore stored fields that are missing from the sentence when making mentions?

It's crashing where we're getting the fields from the index -- which we did that way as an optimization:
https://github.com/lum-ai/odinson/blob/master/core/src/main/scala/ai/lum/odinson/lucene/analysis/TokenStreamUtils.scala#L20

basically, instead of asking each time what fields a doc has, or else getting _all_ the fields, it gets a field that "should" be there.  We can do this diff, but we'd change something here, not in the Mention

marcovzla:
I think the changes to the config provide extra flexibility and may be a good idea to keep them.
But it is true that these changes may caught some people off guard.
Maybe just incrementing the version and being explicit about this in the docs would be enough?

Issue #336, opened 2021-08-30
Data consistency issue  in backend API search results and the stored documents.


curtkohler:
**Describe the bug**
Data consistency issue  in backend API search results and the stored documents.

We have been using our own internal parsing engine(with custom tokenization rules),  formatting the results into Odinson JSON documents, and then loading them into an Odinson index.  While everything works well for the most part, occasionally I've been seeing match spans that appear to be off by one or two. I initially assumed an edge case issue in my code, but digging deeper I think there is a conflict between backend and core. After looking at things in more detail, it became apparent that extra tokens are being generated in the query response's "words" field in at least the execute/pattern API call. As I mentioned, we have a custom tokenization which allows for embedded whitespace in some tokens. These tokens (along with some other cases) appear to be getting split when generating the API response for a query. The api's sentence call results show the index contains the un-split tokens. It also appears that the match offsets returned in the query response align correctly with the un-split tokenization as well.

I believe that the Lucene analyser is being applied to the already parsed tokens when the OdinsonController calls the core's DataGatherer.getTokens() call - which is further spliting things.  That seems odd given that you are already dealing with pre-tokenized fields, but it my be necessary for some other aspect of the code base I haven't looked at.

I'm attaching an example scoreDoc entry from a  query result as well as the corresponding sentence request output. You can see that the token. 'Figure 3' becomes two separate tokens in the query response each time it appears in the sentence.

**Expected behavior**
I would expect Odinson backend API to respect the tokenization of the loaded docs and not re-define it.

[ExampleSentenceResponse.txt](https://github.com/lum-ai/odinson/files/7076936/ExampleSentenceResponse.txt)
[ExampleQueryResponse.txt](https://github.com/lum-ai/odinson/files/7076937/ExampleQueryResponse.txt)

myedibleenso:
Hi, @curtkohler .  Thank you for reporting this.  We'll dig in.  Just to confirm: the sentence in `ExampleSentenceResponse.txt` has the ID 5035?

curtkohler:
Yes, that was returned by the API call:  http://XX.XX.XX.XX:8001/api/sentence?sentenceId=5035. Happy to provide the entire original JSON document that was indexed if needed, but I suspect just creating a test JSON doc with the one sentence (less the two additional fields we've added) would be enough to work with.  Also, there are a couple of other Figure references in that particular sentence that we treat as a single token that are getting split up as well.

danebell:
[...]

PR #337, opened 2021-09-24
Replace `WhitespaceAnalyzer` to allow whitespace within tokens

danebell:
See #336. This update replaces `WhitespaceAnalyzer` with `KeywordAnalyzer`. Whitespace separation should be handled by parsers, not by the `OdinsonIndexWriter`.


BeckySharp (approved):
lgtm thanks!

PR #341, opened 2021-09-29
refactor odinson index operations into interface and provide implementation that supports incremental indexing

reynoldsm88:


reynoldsm88:
I would suggest that, if approved, this PR be "squash merged" so that it is easy to roll back these changes in case any issues are discovered

kwalcock:
Something for us to work on...  Grep for finalize.
```scala
    var searcher: IndexSearcher = null
    try {
      searcher = acquireSearcher()
      searcher.getIndexReader.getTermVectors(docId)
    } catch {
      case e: Throwable => throw new RuntimeException("what is the best way to deal with this?")
    } finally releaseSearcher(searcher)
```

core/src/main/scala/ai/lum/odinson/state/State.scala
```diff
@@ -53,11 +55,25 @@ trait State {
...
+      case "memory" => MemoryState(config)
+      case "mock"   => MockState
+      case _        => throw new Exception(s"Unknown state provider: $provider")
+    }
+
+    state
+  }
+
+  def apply(config: Config, index: OdinsonIndex): State = {
     val provider = config.apply[String]("odinson.state.provider")
```
kwalcock, line 73: It seems like these two could be combined.

reynoldsm88: @kwalcock this may have been a miss on my part, i might have kept the old method there while i implemented the new one, let me take a look.

reynoldsm88:
@kwalcock could you explain your comment about finalize? i'm happy to look into it, but if you know what you are looking for i'd say go ahead and take a look!

kwalcock:
@reynolds88, I have a touch-up PR almost ready for you/us.  I'm not yet quite sure how to integrate the changes, but the code above presently looks like this:
```scala
    usingSearcher { searcher =>
      searcher.getIndexReader.getTermVectors(docId)
    }
```

kwalcock:
Also, LUM should decide "what is the best way to deal with this?"  The OdinsonController also has some questionable error handling, mostly unrelated to this PR, that should be checked and ironed out.

reynoldsm88:
> @reynolds88, I have a touch-up PR almost ready for you/us. I'm not yet quite sure how to integrate the changes, but the code above presently looks like this:
> 
> ```scala
>     usingSearcher { searcher =>
>       searcher.getIndexReader.getTermVectors(docId)
>     }
> ```

@kwalcock sounds good, once you are done push it to this PR and i can take a look...


and i agree that we need to do something about the error handling besides "find the best way to deal with this", I just wasn't sure what the team wanted to do there.

kwalcock:
I see where reynoldsm88:master branched from lum-ai:master and that would be the obvious place to make a dev branch, but there is also one "making sure the indexer is up to date with master" that makes me think that the two branches connect again somewhere else without a trail.  Can @reynolds88 provide a few details?

kwalcock:
I'm probably thinking about that wrong.  I should probably branch off the current master since that's where the PR is headed.

kwalcock:
I'm attempting to move this to #348 so that we can work on it better.  I'll close this one, but reopen if necessary.

reynoldsm88:
@marcovzla sure i'll take a look today

marcovzla:
Thanks! I expect to merge this today.

reynoldsm88:
@danebell whoops sorry, I didn't think that there were really any changes and didn't run the compile and test build before committing 🤦.

I'll push the changes tomorrow morning!

PR #349, opened 2021-10-10
Switch most "finally close" to "using"

kwalcock:


kwalcock:
I'd like to go through all the tests and remove the closes in a different PR.  Also, the "minimize diff" commit should be undone in a subsequent step.  We need to be more careful in the error handling there.  The exception handler should be finished as discussed in  #341.

kwalcock:
@reynoldsm88, you are more than welcome to add comments.  It won't let me make you a reviewer.  These extra PRs are helping us to stage your changes.

kwalcock:
Everyone else, perhaps we can discuss these changes at a Monday meeting.

PR #363, opened 2022-01-15
Remove `ai.lum.odinson.OdinsonIndexWriter`

myedibleenso:
`OdinsonIndexWriter` was refactored and moved by @reynoldsm88 as part of #341 .

from `ai.lum.odinson.` to `ai.lum.odinson.lucene.index.`.  In that process, it seems a change to the analyzer introduced in #337 was inadvertently lost.

Along with some minor cleanup (removing printlns, renaming a test for consistency, etc.), this PR removes `ai.lum.odinson.OdinsonIndexWriter` and updates `ai.lum.odinson.lucene.index.OdinsonIndexWriter` to use `KeywordAnalyzer` (see #337).

PR #364, opened 2022-01-15
Support for deleting and updating indexed Odinson Documents

myedibleenso:
This PR adds support for ... 

1. deleting an `ai.lum.odinson.Document` (and all associated Lucene documents capturing metadata, etc.) from an index via `OdinsonIndex.deleteOdinsonDoc(odinsonDocId)`

2. updating an indexed `ai.lum.odinson.Document` (and all associated Lucene documents capturing metadata, etc.) via `OdinsonIndex.updateOdinsonDoc(doc)`

... and a couple quality of life changes (context managers for the `ExtractorEngine` and `OdinsonIndex`).

~~Lucene Documents for `ai.lum.odinson.Sentence`s and nested metadata fields now include indexed doc IDs, so that all Lucene docs used to represent an Odinson Document are easily retrievable.  **This is a breaking change to the index which will make this version of Odinson incompatible with indexes generated using previous versions of Odinson**, hence the version bump.~~  

~~We've been incrementing the minor version for these kinds of breaking changes, but perhaps we should starting incrementing the major version?~~


myedibleenso:
NOTE: this builds off of changes in https://github.com/lum-ai/odinson/pull/363

core/src/main/scala/ai/lum/odinson/lucene/index/IncrementalOdinsonIndex.scala
```diff
@@ -94,17 +97,20 @@ class IncrementalOdinsonIndex(
         f(closeableSearcherHolder.searcher)
       }
     } catch {
-      case _: Throwable => throw new RuntimeException("what is the best way to deal with this?")
-//      case e: Throwable => {
-//        e.printStackTrace()
-//        throw e
+      case e: Throwable => throw new RuntimeException(s"Error using searcher: ${e.getStackTrace}")
```
kwalcock, line 100: We probably in general need to be more careful with error handling, especially as this moves past research into the product realm.  Scala isn't a big help in this area.  At some point we should look through all the exceptions and devise a more formal strategy.  For example, there might be an OdinsonException which might always be logged somewhere, maybe even to a database or issue tracking system.  The original stack trace, e.getStackTrace, would be a field rather than a string in the message, etc.  If odinson is hosted, we would want to analyze the log files periodically, etc.

core/src/main/scala/ai/lum/odinson/lucene/index/IncrementalOdinsonIndex.scala
```diff
@@ -20,12 +20,15 @@ import org.apache.lucene.index.{
...
+  BooleanQuery,
   Collector,
   CollectorManager,
   IndexSearcher,
-  Query,
   SearcherManager,
-  TopDocs
+  TopDocs,
+  TermQuery,
+  Query
```
kwalcock, line 31: Isn't the formatter supposed to alphabetize these?

core/src/test/scala/ai/lum/odinson/lucene/index/TestIncrementalIndex.scala
```diff
@@ -121,4 +122,46 @@ class TestIncrementalIndex extends OdinsonTest with BeforeAndAfterEach {
...
+    val odinsonDocId = pies.id
+    index.numDocs() shouldBe 4 // # of lucene docs
+    index.luceneDocIdsFor(odinsonDocId).size shouldBe 4
+
+    // Deleting an Odinson document should delete all of its sentences, as well as its metadata (including any nested fields)
+    index.deleteOdinsonDoc(odinsonDocId)
+
+    index.numDocs() shouldBe 0
+    index.luceneDocIdsFor(odinsonDocId).size shouldBe 0
+    index.close()
```
kwalcock, line 141: It doesn't matter much, especially in a test, but it's nice to have things close even if exceptions have been thrown.

version.sbt
```diff
@@ -1 +1 @@
-version in ThisBuild := "0.6.1-SNAPSHOT"
+ThisBuild / version := "0.7.0-SNAPSHOT"
```
kwalcock, line 1: AFAIK, and I just witnessed it yesterday, the sbt release plugin will revert this to "version in ThisBuild" when it runs, so one might skip this instance (or just let it happen naturally).  All the updates elsewhere are appreciated, though.

kwalcock:
I have no idea if it does what it's supposed to do, but the code is very nice and I'm satisfied.

kwalcock (approved):

core/src/main/scala/ai/lum/odinson/lucene/index/IncrementalOdinsonIndex.scala
kwalcock: On second thought, why don't we just throw in e as the cause of the RuntimeException for now.  It can be inspected or logged more easily from there than from the string.

core/src/main/scala/ai/lum/odinson/lucene/index/IncrementalOdinsonIndex.scala
myedibleenso: It's capable of doing that via [`rewrite.imports.sort = scalastyle`](https://scalameta.org/scalafmt/docs/configuration.html#imports-sort--scalastyle), but apparently we haven't included that in the current configuration.  Maybe we should, but that's probably better left for a separate PR.

core/src/main/scala/ai/lum/odinson/lucene/index/IncrementalOdinsonIndex.scala
myedibleenso: Will do.  The old exception was too opaque when I was debugging. : )

core/src/test/scala/ai/lum/odinson/lucene/index/TestIncrementalIndex.scala
myedibleenso: I'll add a context manager to `OdinsonIndex` and override it in `IncrementalOdinsonIndex`.  I hadn't bothered, because I thought the suite was deleting the index dir for each test via the `BeforeAndAfterEach` mixin and `afterEach()` override, but it seems that isn't the case.  I defined something like that for use with the REST API, so I'll just port it over here and make use of it over there after the next release.

version.sbt
myedibleenso: I've actually upgraded the version of the release plugin In this PR.  I wonder if it will still revert...

kwalcock: I *think* the absolute newest is being used in processors now and it does revert there.

build.sbt
```diff
@@ -1,16 +1,24 @@
 import sbtrelease.ReleasePlugin.autoImport.ReleaseTransformations._
 
-organization in ThisBuild := "ai.lum"
+ThisBuild / organization := "ai.lum"
 
-scalaVersion in ThisBuild := "2.12.10"
+ThisBuild / scalaVersion := "2.12.10"
```
kwalcock, line 5: Should we make it a policy to use the highest _._.X version available, like 2.12.15, so that we get the most bug fixes?

core/src/main/scala/ai/lum/odinson/lucene/index/OdinsonIndexWriter.scala
```diff
@@ -154,9 +154,9 @@ class OdinsonIndexWriter(
       OdinsonIndexWriter.PARENT_TYPE,
       Store.NO
     ))
-    metadata.add(new LuceneStoredField(DOC_ID_FIELD, d.id))
+    //metadata.add(new LuceneStoredField(DOC_ID_FIELD, d.id))
     // FIXME: remove
```
kwalcock: I hope someone knows what this fixme is about.  Maybe it's obsolete.

core/src/main/scala/ai/lum/odinson/utils/exceptions/OdinsonException.scala
```diff
@@ -5,14 +5,13 @@ package ai.lum.odinson.utils.exceptions
...
 
 object OdinsonException {
-
-  def defaultMessage(message: String, cause: Throwable) =
-    if (message != null) message
-    else if (cause != null) cause.toString()
-    else null
-
-}
+  def apply(message: String, cause: Throwable): OdinsonException = {
```
kwalcock, line 15: It's not a big deal, but I think this might encourage people to use null as a second argument and it seems like we're trying to avoid those.  Perhaps there could be two apply methods:
```
apply(message: String) = new OdinsonException(message, None)
apply(message: String, cause: Throwable) = new OdinsonException(message, Some(throwable))
```

core/src/main/scala/ai/lum/odinson/lucene/index/OdinsonIndexWriter.scala
myedibleenso: I believe you're looking at an old commit.  That was a leftover from debugging/experimenting.

core/src/main/scala/ai/lum/odinson/utils/exceptions/OdinsonException.scala
myedibleenso: Sure.

PR #366, opened 2022-01-20
Test multiple java versions

myedibleenso:
Test multiple Java versions and distributions as part of CI


kwalcock:
Thanks, @myedibleenso!

PR #371, opened 2022-02-10
Paginate using Lucene Doc ID

myedibleenso:
This PR simplifies the pagination process by using a Lucene doc ID, rather than an `OdinsonScoreDoc`.


myedibleenso:
It looks like we don't have real tests in place for pagination.  We need to add at least a few (ex. `n = 1, after = 1`, `n = 1, after = 3`,  `n = 2, after = 1`, etc).

myedibleenso:
@marcovzla @danebell

PR #392, opened 2023-10-18
Integrate BioNLPProcessor with Odinson

bgyori:
This PR adds the BioNLPProcessor as one of the processor choices used for annotation. See https://github.com/clulab/reach/issues/802.

**WARNING**: review for changing dependencies (this only works if Reach is available as a dependency) and changing default behavior (for illustrative purposes I changed application.conf to choose the BioNLPProcessor). Further changes may be needed to mitigate any unintended consequences of these.

### DIFF
--- a/extra/build.sbt
+++ b/extra/build.sbt
@@ -11,6 +11,7 @@ libraryDependencies ++= {
     "ai.lum"     %% "nxmlreader"            % "0.1.2",
     "org.clulab" %% "processors-main"       % procVersion,
     "org.clulab" %% "processors-corenlp"    % procVersion,
+    "org.clulab" %% "reach-processors"      % "1.6.3-SNAPSHOT"
   )
 
 }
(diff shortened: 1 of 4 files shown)


kwalcock:
Thanks, @bgyori.  I suspect that we might not want to incorporate this into the main branch, but it would be very nice to have the changes around for documentation or to serve as an example.  Maybe we here can incorporate them into a branch for safe keeping.  What does @myedibleenso who knows both reach and odinson think?

extra/build.sbt
```diff
@@ -11,6 +11,7 @@ libraryDependencies ++= {
     "ai.lum"     %% "nxmlreader"            % "0.1.2",
     "org.clulab" %% "processors-main"       % procVersion,
     "org.clulab" %% "processors-corenlp"    % procVersion,
+    "org.clulab" %% "reach-processors"      % "1.6.3-SNAPSHOT"
```
myedibleenso, line 14: I don't think we want to include a snapshot dependency in the default branch.  


@kwalcock , is a version going to be published?  What is the ETA on getting the transformer-based processor integrated into Reach?

kwalcock: Someone ;-) might still verify that the RC version works on aarch.  Someone else might want to prepare a version of breeze/netlib that works on anything other than Linux (https://github.com/luhenry/netlib/issues/21).  I suspect that it will need to be me and that it will require forking that project and will take more time than we want.

